PR #2, opened 2022-04-05
Allow dAPI names to be reset

bbenligiray:
I was in a call with a major use case and they asked if we had an on-chain registry for data feeds that people can generate on-chain assets out of in a permisionless way. We almost already have that with the dAPI names and this PR completes it by:
- Allowing us to reset dAPI names, which allows us to disable names that are not actively used
- And since the above implies that all names that are set to a data feed ID are active, it reverts if the user tries to read using a dAPI name that is not set

### DIFF
--- a/test/dapis/DapiServer.sol.js
+++ b/test/dapis/DapiServer.sol.js
@@ -3076,6 +3076,14 @@ describe('DapiServer', function () {
           expect(beaconSet.timestamp).to.be.equal(timestamp);
         });
       });
+      context('dAPI name not set', function () {
+        it('reverts', async function () {
+          const dapiName = hre.ethers.utils.formatBytes32String('My beacon');
+          await expect(dapiServer.connect(voidSignerAddressZero).readWithDapiName(dapiName)).to.be.revertedWith(
+            'dAPI name not set'
+          );
+        });
+      });
     });
     context('Reader is whitelisted', function () {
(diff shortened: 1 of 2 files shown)


acenolaza (approved):
LGTM 👍🏻

test/dapis/DapiServer.sol.js
```diff
@@ -2918,11 +2918,23 @@ describe('DapiServer', function () {
...
-        it('reverts', async function () {
-          const dapiName = hre.ethers.utils.formatBytes32String('My dAPI');
-          await expect(
-            dapiServer.connect(roles.dapiNameSetter).setDapiName(dapiName, hre.ethers.constants.HashZero)
-          ).to.be.revertedWith('Data feed ID zero');
+        context('Sender is dAPI name setter', function () {
+          it('sets dAPI name', async function () {
+            const dapiName = hre.ethers.utils.formatBytes32String('My dAPI');
+            await dapiServer.connect(roles.dapiNameSetter).setDapiName(dapiName, beaconSetId);
+            await expect(dapiServer.connect(roles.dapiNameSetter).setDapiName(dapiName, hre.ethers.constants.HashZero))
```
acenolaza, line 2925: Maybe here in this same test you could also add another call to `setDapiName` passing some random hash that is not the `HashZero` just to make sure happy path is working in all cases.

PR #3, opened 2022-04-06
Implement a read function at DapiServer that reads a single uint256

bbenligiray:
I have two concerns about the existing read function
- Reading it as a struct and breaking it down into the value and the timestamp may be too difficult for some users
- It will not be compatible with generic oracle interfaces that expect a single 256-bit number

This PR adds two read functions for this purpose

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -685,7 +685,7 @@ contract DapiServer is
     /// @param dataFeedId Data feed ID
     /// @return value Data feed value
     /// @return timestamp Data feed timestamp
-    function readWithDataFeedId(bytes32 dataFeedId)
+    function readDataFeedWithId(bytes32 dataFeedId)
         external
         view
         override
@@ -699,13 +699,13 @@ contract DapiServer is
         return (dataFeed.value, dataFeed.timestamp);
     }
 
-    /// @notice Reads the dAPI with name
+    /// @notice Reads the data feed with dAPI name
     /// @dev The read data feed may belong to a Beacon or dAPI. The reader
     /// must be whitelisted for the hash of the dAPI name.
     /// @param dapiName dAPI name
     /// @return value Data feed value
     /// @return timestamp Data feed timestamp
-    function readWithDapiName(bytes32 dapiName)
+    function readDataFeedWithDapiName(bytes32 dapiName)
         external
         view
         override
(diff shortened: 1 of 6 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -699,13 +699,31 @@ contract DapiServer is
...
+        view
+        override
+        returns (int256 value)
+    {
+        require(
+            readerCanReadDataFeed(dataFeedId, msg.sender),
+            "Sender cannot read"
+        );
+        DataFeed storage dataFeed = dataFeeds[dataFeedId];
+        require(dataFeed.timestamp != 0, "Data feed does not exist");
```
acenolaza, line 716: Do you mind sharing why we now need to validate that the dataFeed exist while reading?

contracts/dapis/DapiServer.sol
```diff
@@ -699,13 +699,31 @@ contract DapiServer is
...
 
-    /// @notice Reads the dAPI with name
+    /// @notice Reads the data feed value with ID
+    /// @param dataFeedId Data feed ID
+    /// @return value Data feed value
+    function readDataFeedValueWithId(bytes32 dataFeedId)
+        external
+        view
+        override
+        returns (int256 value)
```
acenolaza, line 709: Shouldn't this be `int224`? If not, should we explicitly cast `dataFeed.value` to `int256`?

acenolaza (approved):
LGTM with some questions

contracts/dapis/DapiServer.sol
bbenligiray: Normally the reader gets a struct consisting of the value and the timestamp, and is responsible with validating the timestamp (to check if it's fresh enough, etc) Here we're discarding the timestamp, which disables the user from doing that. The least that we can do is to check if the timestamp is 0 because you probably never want to use the value of an uninitialized data feed (which will be 0) in any kind of logic.

There's also an argument to be made towards "you should never half-validate to avoid creating a sense of false security" so this is somewhat subjective. I think the general approach in contract development is to make a best effort in validation though and this is consistent with that.

contracts/dapis/DapiServer.sol
bbenligiray: I initially wanted the interface to look simple but thinking about it a bit more, you're right. One read function returning `int256` while the other returns `int224` implies they are returning different things and can be confusing in its own regard. I'll use `int224`, it doesn't change anything in practice anyway.

PR #4, opened 2022-04-06
Make unlimited reader permanent

bbenligiray:
The unlimited reader was implemented as an AccessControlRegistry role, which meant `manager` could revoke it. In this PR, the unlimited role is implemented in DapiServer in a **un**revokable form, which allows the manager to whitelist a contract without being able to un-whitelist it. This makes the functionality a lot more powerful, but also more dangerous to use (the risk being invalidating the whitelist functionality).

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -45,10 +45,6 @@ contract DapiServer is
         uint32 timestamp;
     }
 
-    /// @notice Unlimited reader role description
-    string public constant override UNLIMITED_READER_ROLE_DESCRIPTION =
-        "Unlimited reader";
-
     /// @notice dAPI name setter role description
     string public constant override DAPI_NAME_SETTER_ROLE_DESCRIPTION =
         "dAPI name setter";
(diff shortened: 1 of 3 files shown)


Ashar2shahid (commented):
> In this PR, the unlimited role is implemented in DapiServer in a revokable form, which allows the manager to whitelist a contract without being able to un-whitelist it.

I think you meant "in an unrevokable form". 

Everything looks good, but I am wondering what the rationale is behind this change, why don't we want to make this revokable ?

bbenligiray:
@Ashar2shahid 
The user can read Beacons or Beacon sets in a trustless way because their IDs are hashes of underlying parameters. However, the user still has to trust `manager` to not unwhitelist them. This limits the use cases we can serve to how much you can trust `manager` (or the extent of the insurance coverage), which means the kind of services we can provide are limited by our marketcap. This PR allows us to set up a data feed using a $100B data provider for a $100B use-case without needing to be trusted at all.

Ashar2shahid (approved):
Got it 👍

PR #7, opened 2022-04-25
Have DapiServer inherit ExtendedMulticall

bbenligiray:
We need to be able to read multiple Beacons/Beacon sets and the block number at that time with a single static call, which is why ExtendedMulticall is needed in DapiServer.

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: MIT
 pragma solidity 0.8.9;
 
+import "../utils/ExtendedMulticall.sol";
 import "../whitelist/WhitelistWithManager.sol";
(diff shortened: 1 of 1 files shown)


Ashar2shahid (approved):
Not related to this change but should Extended multicall be also be an abstract contract because the original multicall is abstract? mainly for the sake of consistency. I remember trail of bits discussion and them pointing this out as well for contracts that are not used by in isolation but the conclusion was that is was not a hard requirement or a pattern.

bbenligiray:
I think that is an anti-pattern because it's used to semantically signal that the contract shouldn't be used on its own but it disables some compile-time checks while doing so, which is dangerous. It's consistently not used in the repo, for example https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/whitelist/Whitelist.sol is not an abstract contract.

PR #20, opened 2022-09-26
Oev sandbox

bbenligiray:


contracts/dapis/DapiServer.sol
```diff
@@ -935,4 +813,163 @@ contract DapiServer is
...
+            "Signature mismatch"
+        );
+        bytes32 beaconId = deriveBeaconId(airnode, templateId);
+        int224 updatedBeaconValue = decodeFulfillmentData(data);
+        require(
+            timestamp > ownedDataFeeds[msg.sender][beaconId].timestamp,
+            "Fulfillment older than Beacon"
+        );
+        // Timestamp validity is already checked by `onlyValidTimestamp`, which
+        // means it will be small enough to be typecast into `uint32`
```
Siegrift, line 846: only until y2106 :D

aquarat:
As per the meeting yesterday, can we add an event to the constructor so we can more easily track these from the operations-database side? Thanks :)

bbenligiray:
Added to https://github.com/api3dao/airnode-protocol-v1/issues/34

PR #21, opened 2022-09-27
Adds tryMulticall function to ExtendedMulticall.sol

### LINKED ISSUE #19: Add a tryMulticall function that allows individual calls to revert
https://github.com/api3dao/airseeker/issues/216

Refer to https://github.com/makerdao/multicall/blob/master/src/Multicall2.sol#L55

acenolaza:
Closes #19


contracts/utils/interfaces/IExtendedMulticall.sol
```diff
@@ -13,4 +13,20 @@ interface IExtendedMulticall is IMulticall {
...
+        bool success;
+        bytes returnData;
+    }
+
+    event FailedDelegatedcall(
+        bytes data,
+        uint256 timestamp,
+        string errorMessage,
+        address indexed sender
+    );
```
acenolaza, line 27: I really dislike this duplication but since `ExtendedMulticall` doesn't implement the `IExtendedMulticall` interface I don't see a way around it. I realized that making the contract implement it makes compilation fail due to `IMulticall` and `Multicall` being separated. I'll see if I can find a way around it

contracts/utils/ExtendedMulticall.sol
```diff
@@ -36,4 +36,40 @@ contract ExtendedMulticall is Multicall {
...
+                data[i]
+            );
+            if (!success) {
+                // We do not bubble up the revert string from `callData`
+                emit FailedDelegatedcall(
+                    data[i],
+                    block.timestamp,
+                    "low-level delegate call failed",
+                    msg.sender
+                );
```
acenolaza, line 68: I don't really know if this event adds any value but I added it to be consistent with [AirnodeProtocol](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/protocol/AirnodeProtocol.sol#L151-L157) (it can be removed if not necessary)

bbenligiray: I would remove

contracts/utils/ExtendedMulticall.sol
```diff
@@ -36,4 +36,40 @@ contract ExtendedMulticall is Multicall {
...
+    event FailedDelegatedcall(
+        bytes data,
+        uint256 timestamp,
+        string errorMessage,
+        address indexed sender
+    );
+
+    function tryMulticall(bytes[] calldata data)
+        external
+        returns (Result[] memory results)
```
bbenligiray, line 54: Return two arrays imo, then you don't need the struct

contracts/utils/ExtendedMulticall.sol
```diff
@@ -36,4 +36,40 @@ contract ExtendedMulticall is Multicall {
...
+    {
+        results = new Result[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
+            (bool success, bytes memory result) = address(this).delegatecall(
```
bbenligiray, line 58: No `verifyCallResultFromTarget()`?

bbenligiray (changes requested):
We can't merge anything that is not deployed yet to main imo. I suppose create a new `dev` branch from main and target that instead.

contracts/utils/ExtendedMulticall.sol
bbenligiray: Related https://github.com/api3dao/airnode-protocol-v1/issues/26

acenolaza:
@bbenligiray I'll keep this PR open until we have the new branch for the DapiServer updates.

contracts/utils/ExtendedMulticall.sol
bbenligiray: Right, got it. Created https://github.com/api3dao/airnode-protocol-v1/issues/28 for later

contracts/utils/ExtendedMulticall.sol
```diff
@@ -36,4 +36,24 @@ contract ExtendedMulticall is Multicall {
...
+        returns (bool[] memory succeeded, bytes[] memory returnData)
+    {
+        succeeded = new bool[](data.length);
+        returnData = new bytes[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
+            (bool success, bytes memory result) = address(this).delegatecall(
+                data[i]
+            );
+            succeeded[i] = success;
+            returnData[i] = result;
```
bbenligiray, line 56:
```suggestion
            (succeeded[i], returnData[i]) = address(this).delegatecall(
                data[i]
            );
```

contracts/utils/ExtendedMulticall.sol
```diff
@@ -36,4 +36,24 @@ contract ExtendedMulticall is Multicall {
...
     }
+
+    /// @notice Receives and executes a batch of function calls on this contract
+    /// @dev Unlike Openzeppelin.multicall(), this function does not revert when a call fails
+    /// @param data Array with each function call data
+    /// @return succeeded Array with each call success condition
+    /// @return returnData Array with each call result
+    function tryMulticall(bytes[] calldata data)
+        external
+        returns (bool[] memory succeeded, bytes[] memory returnData)
```
bbenligiray, line 47: I think the names should be `success` (uncountable) and `returndata` (already plural) to be consistent with https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol#L135 because this is extending the OZ implementation

contracts/utils/ExtendedMulticall.sol
```diff
@@ -36,4 +36,24 @@ contract ExtendedMulticall is Multicall {
     function getBlockBasefee() external view returns (uint256) {
         return block.basefee;
     }
+
+    /// @notice Receives and executes a batch of function calls on this contract
+    /// @dev Unlike Openzeppelin.multicall(), this function does not revert when a call fails
```
bbenligiray, line 41:
```suggestion
    /// @dev Unlike OpenZeppelin's multicall(), this function does not revert when a call fails
```

bbenligiray (approved):
I'm a bit skeptical about adding this into ExtendedMulticall vs. creating another Multicall-variant from it, but we can revisit it later.

PR #32, opened 2022-11-10
Multicall rework

bbenligiray:
Addressing [this](https://github.com/api3dao/airnode-protocol-v1/pull/21#pullrequestreview-1146051724)

OpenZeppelin's Multicall is a bit inefficient because even though it calls `address(this)` it checks if it's a contract. I figured that since we'll need to implement our own `tryMulticall()` and the external versions, may as well implement everything ourselves to get rid of the dependency (at least in the context of Multicall).

I added an external multicall contract in preperation of searcher contracts. Searchers will need to do batched external calls, and possibly "try" these calls, so may as well get this ExternalMulticall out the door and have it audited now.

I replaced the OZ multicalls with the new multicalls, but left the AccessControlRegistry as is (we probably don't want to redeploy them, but perhaps we do?). This means the authorizer contracts will need to be redeployed (but if they are redeployed with the same admin role description the roles will carry over).


contracts/protocol/AirnodeProtocol.sol
```diff
@@ -20,10 +20,10 @@ import "./interfaces/IAirnodeProtocol.sol";
 /// StorageUtils, SponsorshipUtils and WithdrawalUtils also implement some
 /// auxiliary functionality for PSP.
 contract AirnodeProtocol is
+    ExtendedSelfMulticall,
```
bbenligiray, line 23: tryMultical()l at AirnodeProtocol allows batch fulfillment of RRP requests (though that means fulfillments sharing gas limit so it's not very straightforward to implement)

contracts/utils/ExternalMulticall.sol
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./interfaces/IExternalMulticall.sol";
+
+/// @notice Contract that enables calls to external contracts to be batched
+/// @dev Refer to MakerDAO's Multicall.sol for a similar implementation
+contract ExternalMulticall is IExternalMulticall {
```
bbenligiray, line 8: This is essentially the base of a more flexible OwnableCallForwarder to be used by searchers

contracts/utils/ExternalMulticall.sol
```diff
@@ -0,0 +1,69 @@
...
+        bytes[] calldata data
+    ) external override returns (bytes[] memory returndata) {
+        returndata = new bytes[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
+            require(
+                targets[i].code.length > 0,
+                "Multicall target not contract"
+            );
+            bool success;
+            (success, returndata[i]) = targets[i].call(data[i]); // solhint-disable-line avoid-low-level-calls
```
bbenligiray, line 27: Note that it `call()`s and not `delegatecall()`s

contracts/utils/SelfMulticall.sol
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./interfaces/ISelfMulticall.sol";
+
+/// @notice Contract that enables calls to the inheriting contract to be
+/// batched
+/// @dev Refer to OpenZeppelin's Multicall.sol for a similar implementation
+contract SelfMulticall is ISelfMulticall {
```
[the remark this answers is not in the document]
acenolaza: I like the distinction between the 2 contracts by adding Self and External 👍🏻

contracts/utils/ExternalMulticall.sol
acenolaza: 👍🏻 thanks for pointing that out

contracts/utils/SelfMulticall.sol
```diff
@@ -0,0 +1,56 @@
...
+        override
+        returns (bytes[] memory returndata)
+    {
+        returndata = new bytes[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
+            bool success;
+            (success, returndata[i]) = address(this).delegatecall(data[i]); // solhint-disable-line avoid-low-level-calls
+            // Adapted from OpenZeppelin's Address.sol
+            if (!success) {
+                if (returndata.length > 0) {
```
acenolaza, line 27: Shouldn't this be `returndata[i]`?

contracts/utils/SelfMulticall.sol
```diff
@@ -0,0 +1,56 @@
...
+                    // solhint-disable-next-line no-inline-assembly
+                    assembly {
+                        let returndata_size := mload(returndata)
+                        revert(add(32, returndata), returndata_size)
```
acenolaza, line 31: Do you feel it would be to risky trying to decode the revert string from the `returndata`? Like in here https://github.com/boringcrypto/BoringSolidity/blob/master/contracts/BoringBatchable.sol#L28

contracts/utils/SelfMulticall.sol
acenolaza: If yes, then maybe we need a new local memory variable and assign it to returndata[i] right at the end of this function. Something like: 
```
(bool success, bytes memory result) = address(this).delegatecall(data[i]);
if (!success) {
  ...
}
returndata[i] = result;
```

contracts/utils/ExternalMulticall.sol
```diff
@@ -0,0 +1,69 @@
...
+    /// @notice Batches calls to external contracts and reverts if at
+    /// least one of the batched calls reverts
+    /// @dev Bubbles up the raw returndata if a call reverts, leaving decoding
+    /// to the user
+    /// @param targets Array of target addresses of batched calls
+    /// @param data Array of calldata of batched calls
+    /// @param returndata Array of returndata of batched calls
+    function externalMulticall(
+        address[] calldata targets,
+        bytes[] calldata data
```
acenolaza, line 18: Shouldn't we `require` that these 2 have the same `length`?

contracts/utils/ExternalMulticall.sol
```diff
@@ -0,0 +1,69 @@
...
+            }
+        }
+    }
+
+    /// @notice Batches calls to external contracts but does not revert if any
+    /// of the batched calls reverts
+    /// @param targets Array of target addresses of batched calls
+    /// @param data Array of calldata of batched calls
+    /// @param successes Array of success conditions of batched calls
+    /// @param returndata Array of returndata of batched calls
```
acenolaza, line 52: `@param` or `@return`? thinking https://docs.soliditylang.org/en/v0.8.17/natspec-format.html

contracts/protocol/AirnodeProtocol.sol
bbenligiray: Yeah. I suppose it would also have to be limited to requests that specify `_gasLimit` (which doesn't exist) and we would need to add these up for the batch fulfillment

contracts/utils/SelfMulticall.sol
bbenligiray: > Shouldn't this be returndata[i]?

Nice catch

> If yes, then maybe we need a new local memory variable and assign it to returndata[i] right at the end of this function. 

I avoided that because it optimizes the non-happy path at the cost of making the happy path more expensive. In a non-try multicall we don't expect a revert so the non-happy path should be rarer.

contracts/utils/SelfMulticall.sol
bbenligiray: So the Boring implementation assumes `returndata` is a `string` that is ABI-encoded into `bytes`. The OpenZeppelin implementation assumes `returndata` is anything that is ABI-encoded into `bytes`. An additional improvement over that would be not making any assumptions about `returndata`. For example, `returndata` can be `0x1234` (as a result of a future fork), which is not a valid ABI-encoding because the bytelength is not a multiple of 32. The problem is there are a lot of other ways for a `bytes` to not be a valid ABI encoding (and not only because of length), and the only way to verify that is knowing the schema beforehand, which we don't with custom errors (with only string reverts we could have verified it by re-encoding it as in [here](https://github.com/api3dao/airnode-protocol-v1/blob/c7eae72bcd82292a05f7a72d9a81dc7f350974d0/contracts/dapis/DapiServer.sol#L507)).

So the best that can be done is making a best effort as in OZ. The thing to consider is what happens to the OZ implementation if `returndata` stops conforming to ABI (for example if it's `0x1234`). It then misinterprets the first 32 bytes of the `returndata` as the length of the data to be reverted with. In the case of `0x1234`, this will appear as `0x1234..............` (spilling over to the memory space to read the rest), which is a huge number. Then, this is what the call will revert with, and you will pay a lot of gas if you try to emit this as in RRP.

Then to use the OZ implementation, we would have to assume that `returndata` will always conform to contract ABI. The thing is, contract ABI is designed to represent calldata and returndata, so this is somewhat of a safe bet. I think searcher ExternalMulticall contracts should definitely use this, but I'm still not 100% convinced about adding it to AirnodeProtocol and DapiServer

contracts/utils/ExternalMulticall.sol
bbenligiray: Thanks, fixed

contracts/utils/ExternalMulticall.sol
bbenligiray: Added

contracts/utils/SelfMulticall.sol
```diff
@@ -0,0 +1,60 @@
...
+        returns (bytes[] memory returndata)
+    {
+        returndata = new bytes[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
+            bool success;
+            (success, returndata[i]) = address(this).delegatecall(data[i]); // solhint-disable-line avoid-low-level-calls
+            // Adapted from OpenZeppelin's Address.sol
+            if (!success) {
+                bytes memory returndataWithRevertString = returndata[i];
+                if (returndataWithRevertString.length > 0) {
```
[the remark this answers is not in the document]
bbenligiray: Good point, implemented

contracts/utils/mock/MockMulticallTarget.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.9;
+
+contract MockMulticallTarget {
+    error MyError(uint256 fieldAlways123, string fieldAlwaysFoo);
+
+    function alwaysRevertsWithString(int256 argPositive, int256 argNegative)
```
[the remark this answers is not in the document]
bbenligiray: I figured that may imply that the multicall target function needs to be pure and this is the more general modifier. I'll comment that instead.

bbenligiray:
After investigating [this](https://github.com/api3dao/airnode-protocol-v1/pull/32#discussion_r1022567488) a bit more, I realized that the OpenZeppelin implementation doesn't assume that the returndata will ABI-encoded, but rather that the first 32 bits of a `bytes memory` is the length of it. Since this is even a more fundamental assumption, I'm convinced that this is perfectly fine to work with. So DapiServer.sol can inherit SelfMulticall.sol here that bubbles up the revert string. I would keep AirnodeProtocol.sol the same, as [the current implementation](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/protocol/AirnodeProtocol.sol#L327) has its merits (if we bubble up the revert string, the user may not realize that Airnode expected the fulfillment to not revert and think that it blindly attempted it), it's not going to work with custom errors (block explorers etc. won't decode a string field of an event to display the error correctly) and I don't feel confident adapting the assembly implementation.

@acenolaza awaiting your approval to merge

acenolaza (approved):
LGTM 👍🏻

Issue #23, opened 2022-10-04
conditionPspBeaconSetUpdate is not view


bbenligiray:
Have `conditionPspBeaconSetUpdate()` have the `view` modifier

PR #42, opened 2022-11-24
Make PSP condition checking function view

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/23


contracts/dapis/DapiServer.sol
```diff
@@ -710,6 +699,32 @@ contract DapiServer is
...
+        view
+        override
+        returns (int224 value, uint32 timestamp)
+    {
+        uint256 beaconCount = beaconIds.length;
+        require(beaconCount > 1, "Specified less than two Beacons");
+        int256[] memory values = new int256[](beaconCount);
+        uint256 accumulatedTimestamp = 0;
+        for (uint256 ind = 0; ind < beaconCount; ind++) {
+            DataFeed storage dataFeed = dataFeeds[beaconIds[ind]];
```
acenolaza, line 720: This variable seems to me like it could just use the `memory` modifier since we are not going to change the data on the underlying mapping but just use it to read it and store it temporarily. This is a `view` function after all 😄

acenolaza (approved):
LGTM 👍🏻

contracts/dapis/DapiServer.sol
```diff
@@ -468,24 +460,21 @@ contract DapiServer is
         bytes32 subscriptionId, // solhint-disable-line no-unused-vars
         bytes calldata data,
         bytes calldata conditionParameters
-    ) external override returns (bool) {
-        require(msg.sender == address(0), "Sender not zero address");
```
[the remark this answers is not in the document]
bbenligiray: These checks ensure that the function can only be called by a [VoidSigner](https://docs.ethers.io/v5/api/signer/#VoidSigner) because that's the only way `msg.sender` can be `address(0)`. Since VoidSigners can only do static calls, this also means these functions can only be static-called. There are two caveats with this (that we didn't encounter yet): (1) Node implementations of some chains may not support VoidSigner calls (2) `address(0)` can be used to send transactions to implement chain-specific functionality such as roll-ups. So I was never 100% comfortable about this but at least ToB thought it was fine.

The entire reason why we're trying to limit calls to static calls is to prevent people working around whitelisting. For example, if we exposed this contract, other contracts would be able to extract 1 bit of information from the data feed (through the returned `bool`). Since we don't need to worry about this anymore, I removed all those `require`s.

contracts/dapis/DapiServer.sol
bbenligiray: `storage` is the cheaper one, if you use `memory` it gets copied from storage to memory, which will induce the gas overhead

Issue #22, opened 2022-10-04
Add announce functions to StorageUtils


bbenligiray:
Currently there are only template and subscription storage functions, based on the reasoning that the limitations of `getLogs()` is not standard across chains/providers and will change over time. However, it is probably a good idea to have `announceTemplate()` and `announceSubscription()` functions that only emit events and not store anything just in case.

PR #43, opened 2022-11-25
Add template and subscription announcements

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/22


contracts/protocol/StorageUtils.sol
```diff
@@ -49,19 +46,21 @@ contract StorageUtils is IStorageUtils {
     /// @notice Subscription details with the ID
     mapping(bytes32 => Subscription) public override subscriptions;
 
-    /// @notice Stores template details
+    /// @notice Announces template details with an event
+
+    /// @notice Announces template details
```
acenolaza, line 51: This looks like it can be removed:

```suggestion
```

contracts/protocol/StorageUtils.sol
```diff
@@ -70,14 +69,26 @@ contract StorageUtils is IStorageUtils {
...
+    /// @notice Stores template details
+    /// @param endpointId Endpoint ID (allowed to be `bytes32(0)`)
+    /// @param parameters Template parameters, encoded in Airnode ABI
+    /// @return templateId Template ID
+    function storeTemplate(bytes32 endpointId, bytes calldata parameters)
+        external
+        override
+        returns (bytes32 templateId)
+    {
+        templateId = announceTemplate(endpointId, parameters);
```
acenolaza, line 84: Looks kind of weird that we are not emitting the event at the end of the function but I do like re-using the code by calling the announce function 🤷🏻‍♂️

acenolaza (approved):
LGTM 👍🏻

Issue #24, opened 2022-10-04
Consider adding the heartbeat requirements to PSP conditions


bbenligiray:
Currently the Beacon update specs is limited to a deviation threshold. Some reader contracts implement heartbeat requirements (e.g., revert if the data feed hasn't been updated in the last day). Although we do heartbeat updates through Airseeker, this is not a protocol-level requirement, which is not nice.

PR #44, opened 2022-11-25
Add heartbeats to PSP conditions

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/24

I added heartbeats to Beacon/Beacon set update PSP conditions. The condition parameters consist of two fields: deviation threshold and heartbeat interval. If either of these is set to zero, it is ignored. This means one can build a Beacon that doesn't care about the deviation and only updates every X seconds.

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -309,14 +309,16 @@ contract DapiServer is
     ) external view override returns (bool) {
         bytes32 beaconId = subscriptionIdToBeaconId[subscriptionId];
         require(beaconId != bytes32(0), "Subscription not registered");
-        DataFeed storage beacon = dataFeeds[beaconId];
+        // Assuming that the update value will be signed after this condition
+        // returns true, the update timestamp will be larger than
+        // `block.timestamp`, which will still satisfy the update condition.
         return
-            calculateUpdateInPercentage(
-                beacon.value,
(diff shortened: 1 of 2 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -309,14 +309,16 @@ contract DapiServer is
     ) external view override returns (bool) {
         bytes32 beaconId = subscriptionIdToBeaconId[subscriptionId];
         require(beaconId != bytes32(0), "Subscription not registered");
-        DataFeed storage beacon = dataFeeds[beaconId];
+        // Assuming that the update value will be signed after this condition
+        // returns true, the update timestamp will be larger than
+        // `block.timestamp`, which will still satisfy the update condition.
```
acenolaza, line 314: Not sure if this comment adds much value. It seems that we are just saying that the check uses block.timestamp because psp update will happen after this condition check so it will obviously satisfy heartbeat value, right?

acenolaza (approved):
LGTM 👍🏻

contracts/dapis/DapiServer.sol
bbenligiray: Yes. I try to state such assumptions explicitly because otherwise the auditor will point it out and claiming that it's intended behavior afterwards doesn't remove the issue from the audit report.

Issue #37, opened 2022-11-21
Consider using uint224 for data feed values


bbenligiray:
In most cases, data feeds are used for asset prices, which makes all non-positive values invalid. This means all reading contracts should check that the value is positive before using it, which feels wrong. Furthermore, values that float around 0 result in weird deviation % behavior (see [this](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/dapis/DapiServer.sol#L913-L917)). So the current implementation does not actually support signed values in the vanilla form.

The naive solution to this problem is simply using `uint224` as the data feed value. However, we should consider that there are signed kinds of data that we also want to operate data feeds for, and with the current implementation, only changing the data feed value type will mean we won't be able to support these with DapiServer.sol. There is probably a more generalized solution that includes an offset value in the update condition parameters (in addition to the deviation threshold).

PR #46, opened 2022-11-25
Have data feed values be unsigned

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/37

This is obviously superior for price data feeds, whose values are expected to be positive. Also, the signed implementation with %-based deviation thresholds simply didn't work. For example, consider a temperature data feed in Celsius. Whenever the temperature moves between -1 and +1, the data feed is going to update for any reasonable deviation threshold (i.e., that is smaller than 100%), which can't be right.

In the new scheme, the template includes some post-processing (most likely using [airnode-plug](https://github.com/api3dao/airnode/issues/1224)) that will convert the data into a positive number where the deviation threshold makes sense. For example, in the temperature case, we can have airnode-plug specify [273](https://en.wikipedia.org/wiki/Absolute_zero) to be added to the temperature (`.add(273)`) before writing it to the chain. Then, with a deviation threshold of 1%, the data feed will update every 2.73 degree update around 0 degrees Celsius. The user of this data feed will need to subtract 273 before using it in their contract (which needs to be documented off-chain, similar to how we document that we use 18 decimals for our price data feeds).

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -40,10 +40,10 @@ contract DapiServer is
     using ECDSA for bytes32;
 
     // Airnodes serve their fulfillment data along with timestamps. This
-    // contract casts the reported data to `int224` and the timestamp to
+    // contract casts the reported data to `uint224` and the timestamp to
     // `uint32`, which works until year 2106.
     struct DataFeed {
-        int224 value;
+        uint224 value;
         uint32 timestamp;
     }
 
@@ -145,8 +145,8 @@ contract DapiServer is
     /// `setRrpSponsorshipStatus()`), the sponsor must also give update request
     /// permission to the sender (by calling
     /// `setRrpBeaconUpdatePermissionStatus()`) before this method is called.
-    /// The template must specify a single point of data of type `int256` to be
-    /// returned and for it to be small enough to be castable to `int224`
+    /// The template must specify a single point of data of type `uint256` to be
+    /// returned and for it to be small enough to be castable to `uint224`
     /// because this is what `fulfillRrpBeaconUpdate()` expects.
     /// @param airnode Airnode address
     /// @param templateId Template ID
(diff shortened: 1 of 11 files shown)


bbenligiray:
As a note, this requires the `_type` of the existing templates/Beacons to be updated to `uint256`

bbenligiray:
It may also require some modifications to Airseeker

bbenligiray:
@acenolaza in case you missed it, this PR comes before the others

acenolaza:
> @acenolaza in case you missed it, this PR comes before the others

I missed it on purpose since I wanted to spend more time thinking about this changes. I have my reservations with taking the naive approach of switching from int224 to uint224 mostly because:
1. makes handling negative data feed values more complex 
2. doesn't really fix the issue with overly frequent updates when data feed value is close to 0
Am I wrong?

bbenligiray:
> makes handling negative data feed values more complex

My argument is that it was actually not possible to handle signed data feeds even though it looked like it was possible, and this change makes that complex, but at least possible in theory

> doesn't really fix the issue with overly frequent updates when data feed value is close to 0

So the idea is you should always treat your data so that it never gets close to 0. The thing with % deviation threshold is that it has the properties (1) the value needs to have a lower and upper bound for a % to be calculated (2) the absolute value that 1% corresponds to depends on where your value is with respect to your lower and upper bounds. So you should treat/pre-process/normalize your data for the deviation threshold to have reasonable characteristics, this is a more general problem and not related to signed data in particular. So say you want to run a data feed in Celsius for ocean temperature (so you know that it will never be negative), you still need to add an offset value if you're aiming for specific deviation threshold characteristics.

> The idea that feeds are always positive is wrong

That's not what I'm saying

bbenligiray:
I came up with an alternative, I'll open another PR

bbenligiray:
Here's the alternative PR https://github.com/api3dao/airnode-protocol-v1/pull/61 I'm fine with replacing this with the new PR

bbenligiray:
Closed in favor of https://github.com/api3dao/airnode-protocol-v1/pull/61

Issue #38, opened 2022-11-22
Support meta-transactions for authorizers


bbenligiray:
Especially for Airnode-controlled authorizers, we may want to support gasless transactions because Airnode operators/API providers often won't hold the native currency of the chain they're operating on.

PR #52, opened 2022-11-26
Use ERC2771 to enable a limited subset of meta-txes for Authorizers and Allocators

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/38

The main use-case I was thinking about is a user sends a whitelisting request to an Airnode operator on ChainAPI on a random chain. The Airnode operator doesn't have the respective currency, and doesn't want to pay for other people's whitelisting. Then, they create a meta-tx, which ChainAPI sends to the requester to execute. Ideally this should be a whitelisted extension because the Airnode operator cannot undo the whitelisting for the same reasons that they can't execute the initial whitelisting.

Meta-txes typically use nonces to prevent replays. However, considering that there may be multiple parties that request such whitelistings, we can't use nonces. Note that this is because our use-case is fundamentally different from "gasless transactions". With gasless transactions, it's the signer that wants the transaction to be executed. In our case, it is some other party that wants the transactions to be executed, the only reason they can't is because they need the permission of the signer. This enables us to not bother with the complexity of a decentralized relayer network.

### DIFF
--- a/contracts/utils/ExpiringMetaCallForwarder.sol
+++ b/contracts/utils/ExpiringMetaCallForwarder.sol
@@ -1,0 +1,77 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/Address.sol";
+import "@openzeppelin/contracts/utils/cryptography/draft-EIP712.sol";
+import "./SelfMulticall.sol";
+
+/// @notice Contract that forwards expiring meta-calls to ERC2771 contracts
+/// that trust it
+/// @dev The term "meta-call" is used as a subset of "meta-tx" where `value`
+/// is zero and `to` is a contract.
+/// Adapted from OpenZeppelin's MinimalForwarder.sol.
+contract ExpiringMetaCallForwarder is EIP712, SelfMulticall {
+    using ECDSA for bytes32;
+    using Address for address;
+
+    struct ExpiringMetaCall {
+        address from;
+        address to;
+        bytes data;
+        uint256 expirationTimestamp;
+    }
+
+    /// @notice If the meta-call with hash is already executed
+    mapping(bytes32 => bool) public metaCallWithHashIsExecuted;
+
+    bytes32 private constant _TYPEHASH =
+        keccak256(
+            "ExpiringMetaCall(address from,address to,bytes data,uint256 expirationTimestamp)"
+        );
+
+    constructor() EIP712("ExpiringMetaCallForwarder", "1.0.0") {}
+
+    /// @notice Verifies the signature and executes the meta-call
+    /// @dev This implementation is intended for use-cases where both `from`
+    /// and the meta-call executor benefit from the meta-call being executed
+    /// directly (e.g., `from` wants to whitelist the executor but does not
+    /// have funds, so they give the executor a meta-call to execute on their
+    /// behalf). This means a lot of the security considerations that apply in
+    /// general relayer network implementations are irrelevant here.
+    /// @param metaCall Meta-call
+    /// @param signature Meta-call hash signed by `from`
+    /// @return returndata Returndata
+    function execute(
+        ExpiringMetaCall calldata metaCall,
+        bytes calldata signature
+    ) external returns (bytes memory returndata) {
+        bytes32 metaCallHash = _hashTypedDataV4(
+            keccak256(
+                abi.encode(
+                    _TYPEHASH,
+                    metaCall.from,
+                    metaCall.to,
+                    keccak256(metaCall.data),
+                    metaCall.expirationTimestamp
+                )
+            )
+        );
+        require(
+            !metaCallWithHashIsExecuted[metaCallHash],
+            "Meta-call already executed"
+        );
+        require(
+            metaCall.expirationTimestamp > block.timestamp,
+            "Meta-call expired"
+        );
+        require(
+            metaCallHash.recover(signature) == metaCall.from,
+            "Invalid signature"
+        );
+        metaCallWithHashIsExecuted[metaCallHash] = true;
+        returndata = metaCall.to.functionCall(
+            abi.encodePacked(metaCall.data, metaCall.from)
+        );
+    }
+}
--- a/test/authorizers/RequesterAuthorizerWithAirnode.sol.js
+++ b/test/authorizers/RequesterAuthorizerWithAirnode.sol.js
@@ -161,6 +161,9 @@ describe('RequesterAuthorizerWithAirnode', function () {
           expect(await requesterAuthorizerWithAirnode.adminRoleDescription()).to.equal(
             requesterAuthorizerWithAirnodeAdminRoleDescription
           );
+          expect(await requesterAuthorizerWithAirnode.isTrustedForwarder(expiringMetaCallForwarder.address)).to.equal(
+            true
+          );
         });
       });
       context('Admin role description string is empty', function () {
--- a/test/authorizers/RequesterAuthorizerWithAirnode.sol.js
+++ b/test/authorizers/RequesterAuthorizerWithAirnode.sol.js
@@ -304,6 +304,122 @@ describe('RequesterAuthorizerWithAirnode', function () {
         });
       });
     });
+    // Let us demonstrate meta-calls as a proof of concept
+    context('Sender using a meta-call signed by the Airnode address', function () {
+      context('Timestamp extends whitelist expiration', function () {
+        it('extends whitelist expiration', async function () {
+          await accessControlRegistry
+            .connect(airnodeWallet)
+            .renounceRole(whitelistExpirationExtenderRole, airnodeAddress, { gasLimit: 1000000 });
+          let whitelistStatus;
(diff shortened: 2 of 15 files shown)


test/utils/ExpiringMetaCallForwarder.sol.js
```diff
@@ -0,0 +1,275 @@
...
+            const to = expiringMetaCallForwarderTarget.address;
+            const data = expiringMetaCallForwarderTarget.interface.encodeFunctionData('incrementCounter', []);
+            const expirationTimestamp = (await testUtils.getCurrentTimestamp(hre.ethers.provider)) + 3600;
+
+            const domainName = 'ExpiringMetaCallForwarder';
+            const domainVersion = '1.0.0';
+            const domainChainId = (await hre.ethers.provider.getNetwork()).chainId;
+            const domainAddress = expiringMetaCallForwarder.address;
+            /*
+            // Domain separator derivation
```
bbenligiray, line 44: For documentation purposes, I verified that they match the contract values

contracts/utils/ExpiringMetaCallForwarder.sol
```diff
@@ -0,0 +1,77 @@
...
+    /// @notice Verifies the signature and executes the meta-call
+    /// @dev This implementation is intended for use-cases where both `from`
+    /// and the meta-call executor benefit from the meta-call being executed
+    /// directly (e.g., `from` wants to whitelist the executor but does not
+    /// have funds, so they give the executor a meta-call to execute on their
+    /// behalf). This means a lot of the security considerations that apply in
+    /// general relayer network implementations are irrelevant here.
+    /// @param metaCall Meta-call
+    /// @param signature Meta-call hash signed by `from`
+    /// @return returndata Returndata
```
acenolaza, line 44: Maybe something more descriptive like: 
```suggestion
    /// @return returndata Result of the low level call
```

contracts/utils/ExpiringMetaCallForwarder.sol
```diff
@@ -0,0 +1,77 @@
...
+
+    struct ExpiringMetaCall {
+        address from;
+        address to;
+        bytes data;
+        uint256 expirationTimestamp;
+    }
+
+    /// @notice If the meta-call with hash is already executed
+    mapping(bytes32 => bool) public metaCallWithHashIsExecuted;
```
acenolaza, line 26: Why do we need to restrict a meta-call from being executed more than once? To prevent usage on different chains?

contracts/utils/ExpiringMetaCallForwarder.sol
```diff
@@ -0,0 +1,77 @@
...
+            "Meta-call expired"
+        );
+        require(
+            metaCallHash.recover(signature) == metaCall.from,
+            "Invalid signature"
+        );
+        metaCallWithHashIsExecuted[metaCallHash] = true;
+        returndata = metaCall.to.functionCall(
+            abi.encodePacked(metaCall.data, metaCall.from)
+        );
```
acenolaza, line 75: So this is going to work just fine for whitelisting a requester but if this contract is going to be used for other purposes, shouldn't it also forward `value` in the low level call?

It is also really weird to me to see `abi.encodePacked` with a dynamic sized parameter before a fixed sized one (instead of `abi.encode`) but I guess it should be fine because it is like that in the MinimalForwarder.sol code.

One last thing, I guess it will be fine to revert if there is an error in the low level call, right? There will be no need to batch this meta-calls, etc

acenolaza (commented):
Still reviewing but some questions and comments around contracts/utils/ExpiringMetaCallForwarder.sol ...

test/authorizers/RequesterAuthorizerWithAirnode.sol.js
```diff
@@ -148,7 +154,8 @@ describe('RequesterAuthorizerWithAirnode', function () {
           );
           requesterAuthorizerWithAirnode = await requesterAuthorizerWithAirnodeFactory.deploy(
             accessControlRegistry.address,
-            requesterAuthorizerWithAirnodeAdminRoleDescription
+            requesterAuthorizerWithAirnodeAdminRoleDescription,
+            expiringMetaCallForwarder.address
```
acenolaza, line 158: Maybe after 163 we could also expect that [isTrustedForwarder()](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/metatx/ERC2771Context.sol#L20-L22) is equal to this `address`

test/authorizers/RequesterAuthorizerWithAirnode.sol.js
```diff
@@ -292,6 +304,122 @@ describe('RequesterAuthorizerWithAirnode', function () {
...
+            from,
+            to,
+            data,
+            expirationTimestamp: metaCallExpirationTimestamp,
+          };
+          const signature = await airnodeWallet._signTypedData(domain, types, value);
+
+          await expect(expiringMetaCallForwarder.connect(roles.randomPerson).execute(value, signature))
+            .to.emit(requesterAuthorizerWithAirnode, 'ExtendedWhitelistExpiration')
+            .withArgs(airnodeAddress, endpointId, roles.requester.address, airnodeAddress, expirationTimestamp);
```
acenolaza, line 363: 🤯 this is so cool!

The trick here is that _msgSender() == from == airnodeAddress == [account](https://github.com/api3dao/airnode-protocol-v1/blob/meta-tx/contracts/whitelist/WhitelistRolesWithAirnode.sol#L97)

contracts/utils/ExpiringMetaCallForwarder.sol
```diff
@@ -0,0 +1,77 @@
...
+/// Adapted from OpenZeppelin's MinimalForwarder.sol.
+contract ExpiringMetaCallForwarder is EIP712, SelfMulticall {
+    using ECDSA for bytes32;
+    using Address for address;
+
+    struct ExpiringMetaCall {
+        address from;
+        address to;
+        bytes data;
+        uint256 expirationTimestamp;
```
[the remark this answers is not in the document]
bbenligiray: It may potentially reduce the calldata size for `execute()`, I'll check. Depends on if structs in arguments are packed like regular structs or ABI-encoded.

contracts/utils/ExpiringMetaCallForwarder.sol
```diff
@@ -0,0 +1,77 @@
...
+
+    /// @notice If the meta-call with hash is already executed
+    mapping(bytes32 => bool) public metaCallWithHashIsExecuted;
+
+    bytes32 private constant _TYPEHASH =
+        keccak256(
+            "ExpiringMetaCall(address from,address to,bytes data,uint256 expirationTimestamp)"
+        );
+
+    constructor() EIP712("ExpiringMetaCallForwarder", "1.0.0") {}
```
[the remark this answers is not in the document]
bbenligiray: It will be the first version and will go to production as is

contracts/utils/ExpiringMetaCallForwarder.sol
bbenligiray: "Returndata" is more exact imo, it can be argued that "result" is the tuple `(bool success, bytes memory returndata)`

contracts/utils/ExpiringMetaCallForwarder.sol
bbenligiray: Say a token contract is the call forwarding target and someone signed a meta-call to send 10 tokens to someone

contracts/utils/ExpiringMetaCallForwarder.sol
bbenligiray: > So this is going to work just fine for whitelisting a requester but if this contract is going to be used for other purposes, shouldn't it also forward value in the low level call?

Imo it's fine for the same reason why value is excluded in our SelfMulticall/ExternalMulticall contracts

> It is also really weird to me to see abi.encodePacked with a dynamic sized parameter before a fixed sized one (instead of abi.encode) but I guess it should be fine because it is like that in the MinimalForwarder.sol code.
 
 That's an [EIP-2771](https://eips.ethereum.org/EIPS/eip-2771#extracting-the-transaction-signer-address) thing
 
 > One last thing, I guess it will be fine to revert if there is an error in the low level call, right? (This contract inherits SelfMulticall so batching call with the tryMulticall() function should be fine)
 
 Yes, that was my thinking. Maybe I should add a test for it.

test/authorizers/RequesterAuthorizerWithAirnode.sol.js
bbenligiray: Yeah it was surprisingly easy to add. As a note, the `OrIsAirnode` part of `hasWhitelistExpirationExtenderRoleOrIsAirnode` is why I figured we don't need to add meta-tx support to AccessControlRegistry. Not sure if this is entirely valid though, an Airnode operator may still want to create a role and assign it to a contract, and not want to send any transactions themselves.

contracts/utils/ExpiringMetaCallForwarder.sol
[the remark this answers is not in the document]
bbenligiray: I moved `data` after `expirationTimestamp` and ran `test:gas`. The change caused `execute()` to cost ~10-20 more gas. I don't have an explanation :upside_down_face:

contracts/utils/ExpiringMetaCallForwarder.sol
bbenligiray: Added the test

Issue #54, opened 2022-11-26
Enable domain-specific signed data updates to DapiServer


bbenligiray:
Signed data used to update data feeds are replayable across all chain by design. However, we may want to revert this design choice down the path. Consider enabling domain-specific signed data updates as an alternative.

PR #56, opened 2022-11-27
Domain signed data

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/54

I didn't feel like implementing a separate function for this so just allowed both types of signatures, which means using domain-specific signatures will be significantly more expensive (because the contract will need to check if the signature is a generic one first). This may not be a good idea.


bbenligiray:
Closed in favor of https://github.com/api3dao/airnode-protocol-v1/pull/57

Issue #34, opened 2022-11-10
Implement OEV and proxy contracts


bbenligiray:
This includes the changes to the DapiServer.sol and the implementation of the proxy contracts that read dAPIs and Beacon/Beacon sets

bbenligiray:
Emit an event in the constructor for the proxy deployments to be detected easily

Issue #39, opened 2022-11-24
Review implementation


bbenligiray:
After all the issues are addressed, review the implementation with regard to newly introduced code duplication and functions that have now become redundant

bbenligiray:
Largely addressed in https://github.com/api3dao/airnode-protocol-v1/pull/72 but I'm leaving it open for now

PR #57, opened 2022-11-29
Domain signed data v2

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/54

I didn't want to weigh down `updateBeaconSetWithSignedData()` so I just duplicated the implementation for domain-signed data. https://github.com/api3dao/airnode-protocol-v1/issues/34 also requires additions to DapiServer so this may cause the contract size limit to be exceeded. That can be addressed later.

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -717,7 +717,7 @@ contract DapiServer is
     }
 
     /// @notice Updates a Beacon set using data signed for this contract by the
-    ///  respective Airnodes without requiring a request or subscription. The
+    /// respective Airnodes without requiring a request or subscription. The
     /// Beacons for which the signature is omitted will be read from the
     /// storage.
     /// @param airnodes Airnode addresses
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -495,6 +495,40 @@ contract DapiServer is
         emit UpdatedBeaconWithSignedData(beaconId, decodedData, timestamp);
     }
 
+    /// @notice Updates a Beacon using data signed for this contract by the
+    /// respective Airnode, without requiring a request or subscription
+    /// @param airnode Airnode address
+    /// @param templateId Template ID
+    /// @param timestamp Timestamp used in the signature
+    /// @param data Response data (a `uint256` encoded in contract ABI)
+    /// @param signature Template ID, a timestamp and the response data signed
+    /// for this contract by the Airnode address
+    function updateBeaconWithDomainSignedData(
+        address airnode,
+        bytes32 templateId,
+        uint256 timestamp,
+        bytes calldata data,
+        bytes calldata signature
+    ) external override onlyValidTimestamp(timestamp) {
+        require(
+            (
+                keccak256(
+                    abi.encodePacked(
+                        block.chainid,
+                        address(this),
+                        templateId,
+                        timestamp,
+                        data
+                    )
+                ).toEthSignedMessageHash()
+            ).recover(signature) == airnode,
+            "Signature mismatch"
+        );
+        bytes32 beaconId = deriveBeaconId(airnode, templateId);
+        uint256 decodedData = processBeaconUpdate(beaconId, timestamp, data);
+        emit UpdatedBeaconWithSignedData(beaconId, decodedData, timestamp);
+    }
+
     ///                     ~~~PSP Beacon set updates~~~
 
     /// @notice Updates the Beacon set using the current values of its Beacons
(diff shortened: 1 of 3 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -495,6 +495,40 @@ contract DapiServer is
...
+        uint256 timestamp,
+        bytes calldata data,
+        bytes calldata signature
+    ) external override onlyValidTimestamp(timestamp) {
+        require(
+            (
+                keccak256(
+                    abi.encodePacked(
+                        block.chainid,
+                        address(this),
```
Ashar2shahid, line 518: I don't think we should have the contract address be added in the signature, in case we want to upgrade `dapiServer`, it would then require a redeployment across all providers. What is the benefit of having it be appended to the signature?

Ashar2shahid (commented):
didn't take a look at the tests yet

contracts/dapis/DapiServer.sol
bbenligiray: The gateway for this will receive `block.chainid` and `address(this)` as API call parameters and sign accordingly, so it won't require a redeployment. You want the contract address in the signature so that if there is another `DapiServer.sol` (or another contract that uses signatures with the same format) is deployed on the same chain, the signatures for the two contracts can't be replayed on each other.

Ashar2shahid: I didn't know that the gateway received the chainId and address(this), looks good then 👍 . I thought about the signature replay but didn't think it was that much of an issue.

contracts/dapis/DapiServer.sol
```diff
@@ -682,6 +716,94 @@ contract DapiServer is
         );
     }
 
+    /// @notice Updates a Beacon set using data signed for this contract by the
+    ///  respective Airnodes without requiring a request or subscription. The
```
acenolaza, line 720:
```suggestion
    /// respective Airnodes without requiring a request or subscription. The
```

contracts/dapis/DapiServer.sol
```diff
@@ -495,6 +495,40 @@ contract DapiServer is
...
     }
 
+    /// @notice Updates a Beacon using data signed for this contract by the
+    /// respective Airnode, without requiring a request or subscription
+    /// @param airnode Airnode address
+    /// @param templateId Template ID
+    /// @param timestamp Timestamp used in the signature
+    /// @param data Response data (a `uint256` encoded in contract ABI)
+    /// @param signature Template ID, a timestamp and the response data signed
+    /// for this contract by the Airnode address
```
acenolaza, line 505: No need to mention that the `chainId` and the `contract address` is also part of the signed payload?

test/dapis/DapiServer.sol.js
```diff
@@ -302,6 +302,21 @@ describe('DapiServer', function () {
     return [data, signature];
   }
 
+  async function encodeAndDomainSignData(decodedData, templateId, timestamp) {
```
acenolaza, line 305: Maybe the function above (`encodeAndSignData` on line 293) should also use `templateId` instead of `requestHash` as name for the parameter.

contracts/dapis/DapiServer.sol
```diff
@@ -495,6 +495,40 @@ contract DapiServer is
...
+                        timestamp,
+                        data
+                    )
+                ).toEthSignedMessageHash()
+            ).recover(signature) == airnode,
+            "Signature mismatch"
+        );
+        bytes32 beaconId = deriveBeaconId(airnode, templateId);
+        uint256 decodedData = processBeaconUpdate(beaconId, timestamp, data);
+        emit UpdatedBeaconWithSignedData(beaconId, decodedData, timestamp);
```
acenolaza, line 529: Would it be useful in the future to emit a different event than the one emitted in the `updateBeaconWithSignedData()` function?

acenolaza (approved):
LGTM 🚀

contracts/dapis/DapiServer.sol
bbenligiray: That's what I meant by "signed for this contract"

test/dapis/DapiServer.sol.js
bbenligiray: Those functions are a mess, I'll revisit them in https://github.com/api3dao/airnode-protocol-v1/issues/39

contracts/dapis/DapiServer.sol
bbenligiray: I thought about that, but all the infrastructure built to listen to these updates will now also need to be aware of that even though it doesn't matter to them. I suppose it's worth it for the sake of consistency.

Issue #31, opened 2022-11-10
Add (non-OEV) proxy contracts


bbenligiray:
For a more streamlined user flow, we will request users to use proxy contracts with a uniform interface to read data feeds. These contracts need to be implemented.

bbenligiray:
Remove DapiReader as a part of this

bbenligiray:
Only allow data feeds to be read by ID or dAPI name hash (if reading from DapiServer directly) as a part of this

bbenligiray:
Remove "read value" functions as a part of this

PR #60, opened 2022-11-30
Add non-OEV proxy contracts

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/31


contracts/dapis/DapiServer.sol
```diff
@@ -842,69 +844,22 @@ contract DapiServer is
...
         returns (uint224 value, uint32 timestamp)
     {
-        DataFeed storage dataFeed = dataFeeds[dataFeedId];
-        return (dataFeed.value, dataFeed.timestamp);
-    }
-
-    /// @notice Reads the data feed value with ID
-    /// @param dataFeedId Data feed ID
-    /// @return value Data feed value
-    function readDataFeedValueWithId(bytes32 dataFeedId)
```
Ashar2shahid, line 862: Is the assumption that if they want to read directly from the datafeed they have to deploy the OEV proxy ?

Ashar2shahid (approved):
lgtm, just a question

contracts/dapis/DapiServer.sol
bbenligiray: No, `dataFeeds` is `public` now so people can still read from that directly. The reason why there was a `readDataFeedValueWithId()` before is that we needed to check for whitelisting, with that gone `readDataFeedValueWithId()` is obsolete.

contracts/dapis/DapiServer.sol
```diff
@@ -842,69 +844,22 @@ contract DapiServer is
...
     }
 
-    /// @notice Reads the data feed with ID
-    /// @param dataFeedId Data feed ID
+    /// @notice Reads the data feed with dAPI name hash
+    /// @param dapiNameHash dAPI name hash
     /// @return value Data feed value
     /// @return timestamp Data feed timestamp
-    function readDataFeedWithId(bytes32 dataFeedId)
+    function readDataFeedWithDapiNameHash(bytes32 dapiNameHash)
```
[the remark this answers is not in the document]
acenolaza: Actually, I just saw it's being used in the proxy `DapiProxy` contract 🤔 maybe this function would make more sense to be in the caller contract since it seems it will be the only place where it is going to be called from.

contracts/dapis/proxies/DapiProxy.sol
```diff
@@ -0,0 +1,26 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./interfaces/IDapiProxy.sol";
+
```
acenolaza, line 5: Maybe add natspec comments? (on both proxy contacts)

contracts/dapis/proxies/DapiProxy.sol
```diff
@@ -0,0 +1,26 @@
...
+contract DapiProxy is IDapiProxy {
+    address public immutable dapiServer;
+    bytes32 public immutable dapiNameHash;
+
+    constructor(address _dapiServer, bytes32 _dapiName) {
+        dapiServer = _dapiServer;
+        dapiNameHash = keccak256(abi.encodePacked(_dapiName));
```
acenolaza, line 12: Should we emit events in the constructor for apps to collect them (operations-database handler maybe)?

contracts/dapis/proxies/DataFeedProxy.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "./interfaces/IDapiProxy.sol";
+
+contract DataFeedProxy is IDapiProxy {
```
acenolaza, line 6: I wonder if these proxy contracts are good use cases for having a function that calls `selfdestruct()`...it should be safe because these contracts will not hold any funds but it might still need to inherit `onlyOwner`.

contracts/dapis/proxies/interfaces/IDapiProxy.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "../../interfaces/IDapiServer.sol";
+
+interface IDapiProxy {
+    function read() external view returns (uint224 value, uint32 timestamp);
```
acenolaza, line 7: Should we also add the 2 getter functions for the public properties here (`dapiServer`, `dataFeedId`)?

acenolaza (approved):
LGTM 👍🏻 Left a couple of comments and questions

contracts/dapis/DapiServer.sol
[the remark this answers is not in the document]
bbenligiray: This function is meant to save on the number of external calls for dAPI reads. We don't want the proxy contract to have to `call()` DapiServer twice, once to get the data feed ID from the dAPI name (hash) and once more to read the data feed.

contracts/dapis/proxies/DapiProxy.sol
bbenligiray: I seem to have forgotten

contracts/dapis/proxies/DapiProxy.sol
bbenligiray: @Siegrift mentioned that such events can be spoofed (a contract can emit them in a loop, etc.) and be annoying. We need the apps to also know the metadata emitted by the factory contract anyway, so it's better to also emit these at the factory contract and only listen for events from that contract.

contracts/dapis/proxies/DataFeedProxy.sol
bbenligiray: I wouldn't touch `selfdestruct()` with a 10 feet pole atm https://eips.ethereum.org/EIPS/eip-4758

contracts/dapis/proxies/interfaces/IDapiProxy.sol
bbenligiray: Tru

contracts/dapis/proxies/DataFeedProxy.sol
acenolaza: LOL I wasn't aware they are planning on deprecating the opcode 😁 My main motivation to suggest some sort of mechanism to do cleanup if a data feed is ever deprecated. Once we deploy a proxy contract for a data feed then it will be there forever and ever.

contracts/dapis/proxies/DapiProxy.sol
bbenligiray: https://github.com/api3dao/airnode-protocol-v1/issues/65

PR #62, opened 2022-12-02
Implement OEV functionality

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/34

There is extreme code duplication in contracts and tests. I will address that in https://github.com/api3dao/airnode-protocol-v1/issues/39

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -73,6 +73,11 @@ contract DapiServer is
     /// @notice Data feed with ID
     mapping(bytes32 => DataFeed) public override dataFeeds;
 
+    /// @notice Data feed with ID specific to the OEV proxy
+    mapping(address => mapping(bytes32 => DataFeed))
+        public
+        override oevProxyAddressToIdToDataFeed;
+
     /// @notice Data feed ID mapped to the dAPI name hash
     mapping(bytes32 => bytes32) public override dapiNameHashToDataFeedId;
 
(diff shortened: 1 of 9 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -806,6 +811,145 @@ contract DapiServer is
...
+                    abi.encodePacked(templateId, timestamp, data, metadata)
+                ).toEthSignedMessageHash()
+            ).recover(signature) == airnode,
+            "Signature mismatch"
+        );
+        bytes32 beaconId = deriveBeaconId(airnode, templateId);
+        uint224 updatedBeaconValue = decodeFulfillmentData(data);
+        require(
+            timestamp >
+                oevProxyAddressToIdToDataFeed[msg.sender][beaconId].timestamp,
```
Ashar2shahid, line 846: I'm trying to think of a scenario where also checking `dataFeeds[dataFeedId].timestamp` might be relevant but it doesn't seem like it is

Ashar2shahid (commented):
code duplication could possibly be reduced by having a single generic mapping, so `oevProxyAddressToIdToDataFeed` -> `addressToDatafeed`. The global feeds can use `address(beaconSetId)` for the address component. Not sure how we can reduce code duplication in the update functions atm

contracts/dapis/DapiServer.sol
```diff
@@ -806,6 +811,145 @@ contract DapiServer is
...
+                beaconIds[ind] = deriveBeaconId(
+                    airnodes[ind],
+                    templateIds[ind]
+                );
+            } else {
+                bytes32 beaconId = deriveBeaconId(
+                    airnodes[ind],
+                    templateIds[ind]
+                );
+                DataFeed storage dataFeed = dataFeeds[beaconId];
```
bbenligiray, line 926: We read from `dataFeeds[beaconId]` and not `oevProxyAddressToIdToDataFeed[msg.sender][beaconId]`. That's because if a proxy is pointing to a Beacon set, it's constituent Beacons will not receive any OEV updates and `oevProxyAddressToIdToDataFeed[msg.sender][beaconId]` is expected to be empty, while `dataFeeds[beaconId]` may receive individual updates to be used as a fallback.

contracts/dapis/DapiServer.sol
bbenligiray: Nah

acenolaza:
> which you wouldn't want if Beacon ID 1 being faulty was the reason of the switch

Should we maybe add a boolean flag to the `DataFeed` struct in order to prevent updates to "deactivated" data feed? It might require that only an admin can set this variable tho 🤔

bbenligiray:
The current implementation already fully ignores the old data feed as soon as the dAPI is redirected though

Issue #66, opened 2022-12-08
Strip down proxy contracts for cheaper deployment


bbenligiray:
Since the proxy contracts are regularly deployed (and not cloned), we want to minimize their bytecode. Their implementation should be revised accordingly.

bbenligiray:
This issue will be revisited after the refactor

PR #67, opened 2022-12-08
Add ProxyFactory

### LINKED ISSUE #58: Implement the proxy contract factory
We need to be able to detect proxy contract deployments and verify their bytecode. We also need an easy way for the users to deploy new proxy contracts. Both these functionalities can be implemented through a factory. Note that this will be a real factory and not a clone factory so that the resulting proxy makes one less external call / cheaper to use.

bbenligiray:
Closes #58 https://github.com/api3dao/airnode-protocol-v1/issues/65

I started off by implementing what I was talking about earlier, i.e., a factory that can be configured to allow/disallow the deployment of arbitrary contracts. That requires the deployment transaction to provide the entire bytecode in calldata, which costs an additional 20,000 (so 200,000 -> 220,000 gas cost for a DapiProxy deployment, more for the OEV counterparts). I figured that being able to register new proxy contract types isn't worth this overhead, especially since we don't foresee using that feature. So I embedded the four proxy types in the factory contract instead, which results in the cheapest deployment (along with https://github.com/api3dao/airnode-protocol-v1/issues/66).

Note that I also walked back on having people deploy proxies if an identical proxy is already deployed. I think we should discuss how this should be handled again (@RiVal-cz for visibility). It's easy to revert back to the "always deploy a new proxy" scheme, we just replace the Assembly parts with `new`.

### DIFF
--- a/contracts/dapis/proxies/ProxyFactory.sol
+++ b/contracts/dapis/proxies/ProxyFactory.sol
@@ -5,66 +5,64 @@ import "./DataFeedProxy.sol";
 import "./DapiProxy.sol";
 import "./DataFeedProxyWithOev.sol";
 import "./DapiProxyWithOev.sol";
+import "./interfaces/IProxyFactory.sol";
 
-contract ProxyFactory {
-    event DeployedDataFeedProxy(
-        address indexed proxyAddress,
-        bytes32 indexed dataFeedId,
-        bytes metadata
-    );
+/// @title Contract factory that deterministically deploys proxies that read
+/// data feeds (Beacons or Beacon sets) or dAPIs, along with optional OEV
+/// support
+/// @dev The proxies are deployed normally and not cloned to minimize the gas
+/// cost overhead while using them to read data feed values.
+contract ProxyFactory is IProxyFactory {
+    /// @notice DapiServer address
+    address public immutable override dapiServer;
 
-    event DeployedDapiProxy(
-        address indexed proxyAddress,
-        bytes32 indexed dapiName,
-        bytes metadata
-    );
-
-    event DeployedDataFeedProxyWithOev(
-        address indexed proxyAddress,
-        bytes32 indexed dataFeedId,
-        address indexed oevBeneficiary,
-        bytes metadata
-    );
-
-    event DeployedDapiProxyWithOev(
-        address indexed proxyAddress,
-        bytes32 indexed dapiName,
-        address indexed oevBeneficiary,
-        bytes metadata
-    );
-
-    address public immutable dapiServer;
-
+    /// @param _dapiServer DapiServer address
     constructor(address _dapiServer) {
-        require(_dapiServer != address(0), "dAPI server zero");
+        require(_dapiServer != address(0), "DapiServer address zero");
         dapiServer = _dapiServer;
     }
 
+    /// @notice Deterministically deploys a data feed proxy
+    /// @param dataFeedId Data feed ID
+    /// @param metadata Metadata associated with the proxy
     function deployDataFeedProxy(bytes32 dataFeedId, bytes calldata metadata)
         external
+        override
         returns (address proxyAddress)
     {
+        require(dataFeedId != bytes32(0), "Data feed ID zero");
         proxyAddress = address(
             new DataFeedProxy{salt: keccak256(metadata)}(dapiServer, dataFeedId)
         );
         emit DeployedDataFeedProxy(proxyAddress, dataFeedId, metadata);
     }
 
+    /// @notice Deterministically deploys a dAPI proxy
+    /// @param dapiName dAPI name
+    /// @param metadata Metadata associated with the proxy
     function deployDapiProxy(bytes32 dapiName, bytes calldata metadata)
         external
+        override
         returns (address proxyAddress)
     {
+        require(dapiName != bytes32(0), "dAPI name zero");
         proxyAddress = address(
             new DapiProxy{salt: keccak256(metadata)}(dapiServer, dapiName)
         );
         emit DeployedDapiProxy(proxyAddress, dapiName, metadata);
     }
 
+    /// @notice Deterministically deploys a data feed proxy with OEV support
+    /// @param dataFeedId Data feed ID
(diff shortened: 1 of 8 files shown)


contracts/dapis/proxies/ProxyFactory.sol
```diff
@@ -0,0 +1,118 @@
...
+        address indexed oevBeneficiary,
+        bytes metadata
+    );
+
+    address public immutable dapiServer;
+
+    constructor(address _dapiServer) {
+        require(_dapiServer != address(0), "dAPI server zero");
+        dapiServer = _dapiServer;
+    }
```
bbenligiray, line 41: Note that `dapiServer` can also be provided as an argument to the deployment functions, but I figured this is less error-prone

test/dapis/proxies/ProxyFactory.sol.js
```diff
@@ -0,0 +1,101 @@
...
+      )
+    );
+    await hre.ethers.provider.send('evm_setNextBlockTimestamp', [beaconTimestamp + 1]);
+    await dapiServer.updateBeaconWithSignedData(airnodeAddress, templateId, beaconTimestamp, data, signature);
+
+    const proxyFactoryFactory = await hre.ethers.getContractFactory('ProxyFactory', roles.deployer);
+    proxyFactory = await proxyFactoryFactory.deploy(dapiServer.address);
+  });
+
+  describe('deploy', function () {
```
bbenligiray, line 59: Check this out too

test/dapis/proxies/ProxyFactory.sol.js
```diff
@@ -0,0 +1,101 @@
...
+      // Can only deploy once
+      await expect(proxyFactory.deployDataFeedProxy(beaconId, metadata))
+        .to.emit(proxyFactory, 'DeployedDataFeedProxy')
+        .withArgs(proxyAddress, beaconId, metadata);
+      // Subsequent deployments will revert
+      await expect(proxyFactory.deployDataFeedProxy(beaconId, metadata)).to.be.revertedWith('Proxy already deployed');
+
+      // Confirm that the bytecode is the same
+      const dataFeedProxyFactory = await hre.ethers.getContractFactory('DataFeedProxy', roles.deployer);
+      const eoaDeployedDataFeedProxy = await dataFeedProxyFactory.deploy(dapiServer.address, beaconId);
```
bbenligiray, line 87: This costs pretty much the same as calling `deployDataFeedProxy()` minus the gas cost for emitting the event

contracts/dapis/proxies/ProxyFactory.sol
```diff
@@ -0,0 +1,118 @@
...
+        external
+        returns (address proxyAddress)
+    {
+        bytes memory initcode = abi.encodePacked(
+            type(DataFeedProxy).creationCode,
+            abi.encode(dapiServer, dataFeedId)
+        );
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            proxyAddress := create2(0, add(initcode, 0x20), mload(initcode), 0)
```
bbenligiray, line 53: I initially thought the reason deterministic deployments being difficult/impossible on some chains was due to difference in `CREATE2` support/behavior so it should be avoided, but I couldn't find anything about that when I searched for it specifically. So I ended up using it.

contracts/dapis/proxies/ProxyFactory.sol
```diff
@@ -0,0 +1,142 @@
...
+        returns (address proxyAddress)
+    {
+        bytes memory initcode = abi.encodePacked(
+            type(DataFeedProxy).creationCode,
+            abi.encode(dapiServer, dataFeedId)
+        );
+        bytes32 metadataHash = keccak256(metadata);
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            proxyAddress := create2(
```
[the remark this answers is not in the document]
bbenligiray: I thought `create2` was not supported by Solidity yet. I will use that instead.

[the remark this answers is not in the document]
bbenligiray: Done in cbbac2a2c9ca0dba9c873ad09d94694cf3e92fdc. I don't think that a try/catch is necessary, I just didn't want to `require()` without a string.

Siegrift (commented):
@Burak can you remind me why we are using proxies for reading data feeds (even for non oev use cases)? I remember this was discussed few months, but probably on a call… I’d assume in general you would want to have an upgradable proxy because your contract might change which data feed to read from (or even completely change what to read from).

These proxies are immutable, so I am not sure what is the advantage of reading from `IDapiServer` directly. 

Btw. we also want to update https://github.com/api3dao/data-feed-reader-example at some point, right?

bbenligiray:
> can you remind me why we are using proxies for reading data feeds (even for non oev use cases)

The proxies are needed for coverage (to specify the user's policy) and OEV (to determine exact OEV per use-case). It's not necessary if neither of these will be used, but
- People are currently having trouble with reading from the DapiServer directly ([case in point](https://api3workspace.slack.com/archives/C02R79STBRA/p1670848657916399?thread_ts=1670716459.444279&cid=C02R79STBRA)). Furthermore, they are supposed to implement the integration in such a way that they can update it (to point to another dAPI, DapiServer or even oracle solution), which is even more difficult to do correctly. The proxies are useful in that it allows an appropriate integration to be implemented by clicking around on the Market.
- Updating a contract that's designed to read from the DapiServer directly to use a proxy (because it will now use coverage/OEV) will be quite complex and look ugly. When all configurations have the contract read from a proxy with the exact same interface, switching between them becomes trivial (just update the address of the proxy you're reading from).
- We don't want to have our users need to learn two ways of using our product, that means double the amount of guides, examples, etc.

So in general, it's strictly to streamline the UX, which is very important.

> Btw. we also want to update https://github.com/api3dao/data-feed-reader-example at some point, right?

I created https://github.com/api3dao/data-feed-reader-example/issues/7 but I suspect we will also fully revise https://docs.api3.org/dapis/ to be much more Market-focused.

bbenligiray:
I'm merging this and will investigate the clone-based solution in an upcoming PR

PR #72, opened 2022-12-21
Refactor signed data updates

bbenligiray:
Continues from https://github.com/api3dao/airnode-protocol-v1/pull/70, closes https://github.com/api3dao/airnode-protocol-v1/issues/66
Non-OEV proxy deployments cost 185k, and OEV proxy deployments cost 195k gas. This is probably the best we can do in Solidity.

As a part of this, I refactored some signed data updates considering the following:
- The OEV Beacon set updates were hitting the stack size limit. The interface now expects signed data to be ABI-encoded individually to avoid that. For consistency (and also reasons below), signed updates and domain-signed updates are implemented the same way.
- Signed Beacon updates are removed in favor of general signed data feed updates. I did this after confirming that the gas cost improvement of a special signed Beacon update interface to be negligible. I also did many other experiments to reach to the current state of `calldata`/`memory` and `public`/`external` usages. That being said, I sacrificed some gas cost optimization in favor of simplicity and brevity.
- Related to the above two, we'll now expect third parties to do signed data feed updates (both OEV and potentially non-OEV). The previous decoded interface would have been difficult to work with for third parties (they would have hit the same stack size issues and would have to encode the data off-chain and decode it before doing the update).

At the time that I open this PR, OEV signatures include a `signatureCount` to prevent searchers from omitting data with signatures (and refer to the on-chain Beacons instead). Ideally, we would also have a way of preventing them from omitting data without signatures (otherwise they can update a different Beacon set that the dAPI can potentially point to later, this is not a huge issue but it's still not nice). I have a few alternatives in mind, I'll add them before merging. Also, withdrawal tests are missing.


contracts/dapis/DapiServer.sol
```diff
@@ -645,362 +581,265 @@ contract DapiServer is
...
-            msg.sender,
-            updatedValue,
-            updatedTimestamp
-        );
+    }
+
+    /// @notice Withdraws the balance of the OEV proxy to the respective
+    /// beneficiary account
+    /// @param oevProxy OEV proxy
+    function withdraw(address oevProxy) external override {
```
acenolaza, line 833: Let me double check I got this right:
1. anyone can trigger the transfer
2. the balance to be transferred is the sum of all the bid amounts the winning searcher has sent when doing updates
3. the oevBeneficiary is set in the proxy constructor with an address, this address is probably set by us in the factory contract. right? These are API3 proceeds for granting the update to a searcher.

acenolaza (commented):
This refactor looks really nice but is quite challenging to follow due to the diff UI jumping from one line to another 😁

contracts/dapis/DapiServer.sol
```diff
@@ -645,362 +581,265 @@ contract DapiServer is
...
+                    uint32 beaconTimestamp
+                ) = decodeSignedData(signedData[ind]);
                 beaconIds[ind] = beaconId;
+                if (beaconTimestamp != 0) {
+                    values[ind] = beaconValue;
+                    accumulatedTimestamp += beaconTimestamp;
+                } else {
+                    DataFeed storage beacon = dataFeeds[beaconId];
+                    values[ind] = beacon.value;
+                    accumulatedTimestamp += beacon.timestamp;
```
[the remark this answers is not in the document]
bbenligiray: `decodeSignedData` returns a zero timestamp if `data` and `signature` are empty

contracts/dapis/DapiServer.sol
bbenligiray: 1. Yes (though I accept that this is ugly). The reasoning is that the beneficiary may be something like a DAO treasury and may not be able to trigger the withdrawal itself. We also don't want to forward the transfer to the beneficiary in case the beneficiary default `payable` function is heavy (pull payments > push payments).
2. It's the sum of all paid bids to do updates for a specific proxy (the funds may come from different searchers). Funds are collected per proxy and not per beneficiary because the beneficiary may want to implement a proxy-specific incentivization scheme (for example there is a LP pool for each asset and the OEV collected from each asset is awarded to the respective pool -- though it can be argued that the beneficiary addresses will be different then).
3. The project that will deploy the proxy specifies `oevBeneficiary`. They enter that in a textbox on the Market and click deploy, we have them deploy their proxy with this address. Then, they can withdraw the OEV to the address whenever they want.

bbenligiray:
As mentioned above, `signatureCount` prevents the searcher from omitting signed data for a more favorable aggregation. However, there are still two methods of manipulation
1. The searcher can mix and match signed data from two consecutively won auctions for a more favorable aggregation/series of updates that generate artificial OEV opportunities
2. The searcher can omit unsigned data to update another Beacon set, which the respective DapiProxyWithOev may later be pointed to
Both of these are mild cases but it's best to nip them in the bud. For (1), I will add an `updateId` to the metadata that will be signed. This will be a UUID of type `bytes32` generated by the OEV Relay. This will both be used to prevent signed data from different updates to be used together and the OEV Relay to track if updates are executed (so this will be emitted as an indexed log field). For (2), `signedData.length` will be added to the metadata that will be signed.

PR #74, opened 2022-12-26
Use unchecked wherever possible

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/73

I commented my assumptions, the others are protected by types/logic


contracts/dapis/Median.sol
```diff
@@ -21,18 +21,48 @@ contract Median is Sort, Quickselect {
...
+            // No risk of overflow if the signs are different, add them and
+            // divide by 2
+            unchecked {
+                return (x + y) / 2;
+            }
+        } else {
+            // There is risk of overflow if the signs are the same, divide by 2
+            // before adding and compensate for the gobbled bit
+            unchecked {
+                return x / 2 + y / 2 + ((x % 2) + (y % 2)) / 2;
```
bbenligiray, line 65: It feels like this can be implemented without any `if`s

contracts/protocol/WithdrawalUtils.sol
```diff
@@ -43,14 +43,20 @@ contract WithdrawalUtils is IWithdrawalUtils {
...
+            withdrawalRequestId = keccak256(
+                abi.encodePacked(
+                    block.chainid,
+                    address(this),
+                    msg.sender,
+                    ++sponsorToWithdrawalRequestCount[msg.sender]
+                )
+            );
+        }
+
```
acenolaza, line 59: Does it really make sense to use `unchecked` in places like this where we are not inside a loop and the gas savings on executing this tx with `unchecked` is less that 200 gas if I'm not mistaken?

contracts/dapis/Median.sol
```diff
@@ -21,18 +21,48 @@ contract Median is Sort, Quickselect {
...
+            }
+        }
+    }
+
+    /// @notice Averages two signed integers without overflowing
+    /// @param x Integer x
+    /// @param y Integer y
+    /// @return Average of integers x and y
+    function average(int256 x, int256 y) private pure returns (int256) {
+        if (x > 0 != y > 0) {
```
acenolaza, line 55: I get the feeling that a negated XOR could be used here somehow but it might end up being harder to read.

contracts/dapis/Median.sol
```diff
@@ -21,18 +21,48 @@ contract Median is Sort, Quickselect {
...
+            // No risk of overflow if the signs are different, add them and
+            // divide by 2
+            unchecked {
+                return (x + y) / 2;
```
acenolaza, line 59: It seems that this is safer: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/math/Math.sol#L36

contracts/dapis/Median.sol
acenolaza: Can you please add an example to explain how this code is safe to use with `unchecked`?

bbenligiray: The sum only overflows if the added numbers have the same sign. So we can safely add and divide by 2 if they don't have the same sign. If they do have the same sign, we now have to divide by 2 before adding the numbers so that the sum doesn't overflow. However, doing so gobbles the carry bit of the summation that two odd numbers would cause, which we have to compensate for.
If we check the boundary conditions, say x = 2^255-1, y = 2^255-1 (largest positive signed integers)
x / 2 = 2^254 - 1
y / 2 = 2^254 - 1
((x % 2) + (y % 2)) / 2 = (1 + 1) / 2 = 1
x / 2 + y / 2 + ((x % 2) + (y % 2)) / 2 = 2^255-1, both doesn't overflow and is correct.
Say x = -2^255, y = -2^255 (smallest negative signed integers)
x / 2 = -2^254
y / 2 = -2^254
((x % 2) + (y % 2)) / 2 = (0 + 0) / 2 = 0
x / 2 + y / 2 + ((x % 2) + (y % 2)) / 2 = -2^255 both doesn't underflow and is correct.

contracts/protocol/WithdrawalUtils.sol
bbenligiray: I think it's even less, I mainly did it for the sake of consistency. Perhaps they can be removed for the request counts.

contracts/dapis/Median.sol
bbenligiray: Unfortunately that doesn't work with signed integers

| x  | y | result |
| ------------- | ------------- |  ------------- |
| 2  | 1  | 1 |
| 2  | 2  | 2 |
| 1  | 1  | 1 |
| -2  | -1  | -2 (!) |
| -1 | -1  | -1 |
| -2  | -2  | -2 |
| 2  | -1  | 1 (!) |
| 2  | -2  | 0 |
| 1  | -1  | 0 |
| 1  | -2  | 0 |

contracts/dapis/Median.sol
```diff
@@ -21,19 +21,48 @@ contract Median is Sort, Quickselect {
...
+                return average(array[mid1], array[mid2]);
             }
         }
     }
+
+    /// @notice Averages two signed integers without overflowing
+    /// @param x Integer x
+    /// @param y Integer y
+    /// @return Average of integers x and y
+    function average(int256 x, int256 y) private pure returns (int256) {
```
bbenligiray, line 54: This now emulates `(x + y) / 2` with no overflow/underflow risk while only using the cheapest operations. See the tests.

| x  | y | result |
| ------------- | ------------- |  ------------- |
| 2  | 1  | 1 |
| 2  | 2  | 2 |
| 1  | 1  | 1 |
| -2  | -1  | -1 |
| -1 | -1  | -1 |
| -2  | -2  | -2 |
| 2  | -1  | 0 |
| 2  | -2  | 0 |
| 1  | -1  | 0 |
| 1  | -2  | 0 |

contracts/dapis/Median.sol
bbenligiray: `if` is removed in https://github.com/api3dao/airnode-protocol-v1/pull/74/commits/00a296d4c9e7896217d14e33dea8bbd5bd1b6bc9

contracts/dapis/Median.sol
```diff
@@ -21,19 +21,50 @@ contract Median is Sort, Quickselect {
...
+                (y >> 1) +
+                (x & y & 1);
+            // If the average rounded down to negative infinity is negative
+            // (i.e., its 256th sign bit is set), and one of (x, y) is even and
+            // the other one is odd (i.e., the 1st bit of their xor is set),
+            // add 1 to round the average down to zero instead
+            return
+                averageRoundedDownToNegativeInfinity +
+                (int256(
+                    (uint256(averageRoundedDownToNegativeInfinity) >> 255)
```
bbenligiray, line 66: Solidity doesn't give us logical shift (`>>>`) and forces us to use arithmetic shift. Here, we cast the number to unsigned, do an arithmetic shift (which is the same thing as a logical shift for unsigned numbers) and back to a signed integer. So `int256(uint256(signedNumber) >> 255)` is the same thing as `signedNumber >>> 255`, it will be 1 if `signedNumber` is negative and 0 otherwise.

contracts/dapis/Median.sol
```diff
@@ -21,19 +21,50 @@ contract Median is Sort, Quickselect {
...
+                ) & (x ^ y));
```
bbenligiray, line 67: `x ^ y` is odd if (x, y) is a pair of even and odd numbers, this applies whatever their signs are

contracts/dapis/Median.sol
```diff
@@ -21,19 +21,50 @@ contract Median is Sort, Quickselect {
...
         }
     }
+
+    /// @notice Averages two signed integers without overflowing
+    /// @param x Integer x
+    /// @param y Integer y
+    /// @return Average of integers x and y
+    function average(int256 x, int256 y) private pure returns (int256) {
+        unchecked {
+            int256 averageRoundedDownToNegativeInfinity = (x >> 1) +
```
bbenligiray, line 56: Funny thing is Solidity used to round signed numbers down to zero (https://docs.soliditylang.org/fr/latest/types.html#shifts) but they changed it in 0.5.0. In C it's implementation dependent (https://stackoverflow.com/questions/1857928/right-shifting-negative-numbers-in-c), but I think it's more typical to round down to negative infinity.

contracts/protocol/WithdrawalUtils.sol
bbenligiray: I removed the `unchecked`s that require assumptions (except the timestamp accumulation because it happens in a loop)

acenolaza (approved):
Everything in Median.sol is really cleaver 🤯

Issue #59, opened 2022-11-30
Remove endpoint-specific authorization


bbenligiray:
It seems like we will always need authorizations to be airnode-wide/non-endpoint specific. We can then remove endpoint-specific authorization functionality.

PR #77, opened 2023-01-02
Remove endpoint-specific authorizations

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/59


contracts/authorizers/RequesterAuthorizer.sol
```diff
@@ -5,14 +5,11 @@ import "@openzeppelin/contracts/metatx/ERC2771Context.sol";
...
-/// temporarily or permanently authorizes requesters for Airnode–endpoint pairs
-/// or Airnodes
-/// @dev An authorization for an Airnode with endpoint ID `bytes32(0)`
-/// represents a blanket authorization across all endpoints of the Airnode
+/// temporarily or permanently authorizes requesters for Airnodes
 abstract contract RequesterAuthorizer is ERC2771Context, IRequesterAuthorizer {
     struct AuthorizationStatus {
-        uint64 expirationTimestamp;
-        uint192 indefiniteAuthorizationCount;
+        uint32 expirationTimestamp;
```
bbenligiray, line 11: Using 32 bit timestamps for consistency

contracts/authorizers/RequesterAuthorizer.sol
```diff
@@ -234,114 +217,16 @@ abstract contract RequesterAuthorizer is ERC2771Context, IRequesterAuthorizer {
...
-    /// @dev This method has redundant arguments because V0 authorizer
-    /// contracts have to have the same interface and potential authorizer
-    /// contracts may require to access the arguments that are redundant here
-    /// @param requestId Request ID
-    /// @param airnode Airnode address
-    /// @param endpointId Endpoint ID
-    /// @param sponsor Sponsor address
-    /// @param requester Requester address
-    /// @return Authorization status of the request
-    function isAuthorizedV0(
```
bbenligiray, line 260: We now have authorizer types specified in [`config.chains.authorizers`](https://docs.api3.org/airnode/v0.9/reference/deployment-files/config-json.html#authorizers) so even if we wanted to use this with a V0 Airnode we wouldn't need it to satisfy this interface

contracts/authorizers/RequesterAuthorizer.sol
acenolaza: Should we also consider doing the same for `expirationTimestamp` in Allocator contracts? https://github.com/api3dao/airnode-protocol-v1/blob/remove-endpoint-authorization/contracts/allocators/interfaces/IAllocator.sol#L43

acenolaza (commented):
Much simpler LGTM 👍🏻

acenolaza (approved):
Much simpler LGTM 👍🏻

contracts/authorizers/RequesterAuthorizer.sol
bbenligiray: Will do

PR #81, opened 2023-01-05
Have AccessControlRegistry be the trusted forwarder

bbenligiray:
@acenolaza 's comment about fetching `api3Token` from the StakeableVesting implementation made me think that we can fetch the trusted forwarder address from AccessControlRegistry so that it's not possible to mess things up during deployment. After that, I figured that it would be best if AccessControlRegistry's constructor deployed the trusted forwarder so even it cannot refer to the wrong address. Then I realized that AccessControlRegistry itself needs to support meta-txes, which requires us to know the trusted forwarder address before it's constructor begins (because the address needs to go in the ERC2771Context constructor). Finally, I figured that the neatest solution is for AccessControlRegistry to be the trusted forwarder for both itself and all the AccessControlRegistryAdminned contracts.
~This implies that all AccessControlRegistryAdminned contracts are meta-tx enabled, but it would also be possible to use AccessControlRegistry without supporting meta-txes by creating a custom version of AccessControlRegistryAdminned that doesn't inherit ERC2771Context, then the AccessControl Registry would no longer be able to forward meta-txes to the adminned contract.~ AccessControlRegistryAdminned are now not meta-tx enabled by default, it needs to be enabled explicitly by inheriting ERC2771Context and setting AccessControlRegistry to be a trusted forwarder.


acenolaza:
> @acenolaza 's comment about fetching `api3Token` from the StakeableVesting implementation made me think that we can fetch the trusted forwarder address from AccessControlRegistry so that it's not possible to mess things up during deployment. After that, I figured that it would be best if AccessControlRegistry's constructor deployed the trusted forwarder so even it cannot refer to the wrong address. Then I realized that AccessControlRegistry itself needs to support meta-txes, which requires us to know the trusted forwarder address before it's constructor begins (because the address needs to go in the ERC2771Context constructor). Finally, I figured that the neatest solution is for AccessControlRegistry to be the trusted forwarder for both itself and all the AccessControlRegistryAdminned contracts. This implies that all AccessControlRegistryAdminned contracts are meta-tx enabled, but it would also be possible to use AccessControlRegistry without supporting meta-txes by creating a custom version of AccessControlRegistryAdminned that doesn't inherit ERC2771Context, then the AccessControl Registry would no longer be able to forward meta-txes to the adminned contract.

This indeed looks really neat and clever but I'm now getting confused and I feel like AccessControlRegistry is missing an `execute()` function that validates the signature similar to what [ExpiringMetaTxForwarder](https://github.com/api3dao/airnode-protocol-v1/blob/meta-tx-forward-by-acr/contracts/utils/ExpiringMetaTxForwarder.sol#L50) does. Or do we just rely on the EOA submitting the tx to have the proper role granted?

bbenligiray:
AccessControlRegistry does have an `execute()` function because it inherits ExpiringMetaTxForwarder https://github.com/api3dao/airnode-protocol-v1/blob/d437a0d20627ee8a1ba7110c144198bb1c15c9bd/contracts/access-control-registry/AccessControlRegistry.sol#L23

bbenligiray:
I figured all AccessControlRegistryAdminned supporting meta-tx is a bad idea so I now require it to be enabled explicitly. Meta-tx support makes `msg.sender` slightly more heavier and also it's not clear how it will interact with the Airnode protocol, so I didn't want it in DapiServer.

acenolaza:
> AccessControlRegistry does have an `execute()` function because it inherits ExpiringMetaTxForwarder
> 
> https://github.com/api3dao/airnode-protocol-v1/blob/d437a0d20627ee8a1ba7110c144198bb1c15c9bd/contracts/access-control-registry/AccessControlRegistry.sol#L23

LOL, I need to stop doing code reviews at the end of the work day 🤦🏻 😁

acenolaza (approved):
LGTM 👍🏻

PR #82, opened 2023-01-06
Add meta-tx nullification

bbenligiray:
I figured it's a good idea to be able to nullify previously issued signatures. The nullification can also be done through meta-txes if the inheriting contract supports it (which AccessControlRegistry does).

### DIFF
--- a/contracts/utils/ExpiringMetaTxForwarder.sol
+++ b/contracts/utils/ExpiringMetaTxForwarder.sol
@@ -11,7 +11,7 @@ import "./interfaces/IExpiringMetaTxForwarder.sol";
 /// trust it
 /// @notice `msg.value` is not supported. Identical meta-txes are not
 /// supported. Target account must be a contract. Signer of the meta-tx is
-/// allowed to nullify it before execution, effectively rendering the signature
+/// allowed to cancel it before execution, effectively rendering the signature
 /// useless.
 /// This implementation is not intended to be used with general-purpose relayer
 /// networks. Instead, it is meant for use-cases where the relayer wants the
@@ -28,12 +28,10 @@ contract ExpiringMetaTxForwarder is Context, EIP712, IExpiringMetaTxForwarder {
     using ECDSA for bytes32;
     using Address for address;
 
-    /// @notice If the meta-tx with hash is executed or nullified
+    /// @notice If the meta-tx with hash is executed or canceled
     /// @dev We track this on a meta-tx basis and not by using nonces to avoid
     /// requiring users keep track of nonces
-    mapping(bytes32 => bool)
-        public
-        override metaTxWithHashIsExecutedOrNullified;
+    mapping(bytes32 => bool) public override metaTxWithHashIsExecutedOrCanceled;
 
     bytes32 private constant _TYPEHASH =
         keccak256(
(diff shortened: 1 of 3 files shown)


acenolaza (approved):
Changes look good 👍🏻 

Nit: this is just personal preference but I would remove `nullified` references in the code and rename the function `nullify()` to `cancel()` where the `metaTxHash` is set to `true`. To me adding `nullify` implies that something is being set to `null` and makes everything harder to read.

bbenligiray:
I had my doubts about the terminology, I will look for an alternative

contracts/utils/ExpiringMetaTxForwarder.sol
```diff
@@ -22,19 +24,16 @@ import "./interfaces/IExpiringMetaTxForwarder.sol";
...
-    /// using a nonce that is specific to the signer–relayer pair, yet this is
-    /// decided against in favor of the simple UX that not having any kind of a
-    /// nonce provides.
-    mapping(bytes32 => bool) public override metaTxWithHashIsExecuted;
+    /// @notice If the meta-tx with hash is executed or nullified
+    /// @dev We track this on a meta-tx basis and not by using nonces to avoid
+    /// requiring users keep track of nonces
+    mapping(bytes32 => bool)
+        public
+        override metaTxWithHashIsExecutedOrNullified;
```
Ashar2shahid, line 36: maybe change to `metaTxWithHashIsProcessed`

contracts/utils/ExpiringMetaTxForwarder.sol
```diff
@@ -51,7 +50,35 @@ contract ExpiringMetaTxForwarder is EIP712, IExpiringMetaTxForwarder {
...
+            abi.encodePacked(metaTx.data, metaTx.from)
+        );
+    }
+
+    /// @notice Called by a meta-tx source to prevent it from being executed
+    /// @dev This can be used to nullify meta-txes that were issued
+    /// accidentally, e.g., with an unreasonably large expiration timestamp,
+    /// which may create a dangling liability
+    /// @param metaTx Meta-tx
+    function nullify(ExpiringMetaTx calldata metaTx) external override {
```
Ashar2shahid, line 69: agree with @acenolaza should be changed to something like `cancelMetaTx()`

Ashar2shahid (approved):

contracts/utils/ExpiringMetaTxForwarder.sol
bbenligiray: I will use `cancel()` similar to `execute()`. The argument is a part of the function signature so it's clear that it's a meta-tx that's being canceled.

contracts/utils/ExpiringMetaTxForwarder.sol
bbenligiray: I think that's unnecessarily vague (in that what processing means is not clear). It could be `metaTxWithHashIsExecutable` but that implies that the meta tx with hash exists, which it may not, so this is the best option.

bbenligiray:
I initially thought cancel is not good because it may imply that already-executed transactions can also be canceled, but transactions canceling is already [a thing](https://metamask.zendesk.com/hc/en-us/articles/360015489251-How-to-speed-up-or-cancel-a-pending-transaction) so this shouldn't be an issue.

Issue #78, opened 2023-01-03
Use uint32 for timestamps in allocator contracts


bbenligiray:
https://github.com/api3dao/airnode-protocol-v1/pull/77#discussion_r1060142824

PR #84, opened 2023-01-07
Refactor allocators/ and authorizers/

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/78

It looks like I'm running out of time so I decided not to refactor the unit tests comprehensively

### DIFF
--- a/test/allocators/AllocatorWithAirnode.sol.js
+++ b/test/allocators/AllocatorWithAirnode.sol.js
@@ -358,13 +358,13 @@ describe('AllocatorWithAirnode', function () {
     });
   });
 
-  describe('setterOfSlotIsCanStillSet', function () {
+  describe('setterOfSlotCanStillSet', function () {
     context('Setter of slot is still a slot setter', function () {
       it('returns true', async function () {
         await allocatorWithAirnode
           .connect(roles.slotSetter)
           .setSlot(roles.airnode.address, slotIndex, subscriptionId, expirationTimestamp);
-        expect(await allocatorWithAirnode.setterOfSlotIsCanStillSet(roles.airnode.address, slotIndex)).to.equal(true);
+        expect(await allocatorWithAirnode.setterOfSlotCanStillSet(roles.airnode.address, slotIndex)).to.equal(true);
       });
     });
     context('Setter of slot is the Airnode address', function () {
@@ -372,7 +372,7 @@ describe('AllocatorWithAirnode', function () {
         await allocatorWithAirnode
           .connect(roles.airnode)
           .setSlot(roles.airnode.address, slotIndex, subscriptionId, expirationTimestamp);
-        expect(await allocatorWithAirnode.setterOfSlotIsCanStillSet(roles.airnode.address, slotIndex)).to.equal(true);
+        expect(await allocatorWithAirnode.setterOfSlotCanStillSet(roles.airnode.address, slotIndex)).to.equal(true);
       });
     });
     context('Setter of slot is no longer authorized', function () {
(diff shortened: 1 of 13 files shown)


bbenligiray:
Oops I forgot to check out a new branch before pushing `authorizers/`, though I suppose this is fine because the work is very similar

test/allocators/AllocatorWithAirnode.sol.js
```diff
@@ -1,36 +1,30 @@
...
+    const accessControlRegistry = await accessControlRegistryFactory.deploy();
+    const allocatorWithAirnodeFactory = await ethers.getContractFactory('AllocatorWithAirnode', roles.deployer);
+    const allocatorWithAirnode = await allocatorWithAirnodeFactory.deploy(
       accessControlRegistry.address,
       allocatorWithAirnodeAdminRoleDescription
     );
     const airnodeRootRole = testUtils.deriveRootRole(roles.airnode.address);
-    const airnodeAdminRole = await allocatorWithAirnode.deriveAdminRole(roles.airnode.address);
-    airnodeSlotSetterRole = await allocatorWithAirnode.deriveSlotSetterRole(roles.airnode.address);
+    const airnodeAdminRole = testUtils.deriveRole(airnodeRootRole, allocatorWithAirnodeAdminRoleDescription);
```
acenolaza, line 26: (Kind of related to this change) Can you please remind me why AllocatorWithAirnode.sol does not call `_deriveAdminRole()` in [AccessControlRegistryAdminned.sol](https://github.com/api3dao/airnode-protocol-v1/blob/refactor-allocators/contracts/access-control-registry/AccessControlRegistryAdminned.sol#L51) like [RequesterAuthorizerWithAirnode.sol](https://github.com/api3dao/airnode-protocol-v1/blob/refactor-allocators/contracts/authorizers/RequesterAuthorizerWithAirnode.sol#L111) does?

test/allocators/AllocatorWithManager.sol.js
```diff
@@ -363,54 +456,122 @@ describe('AllocatorWithManager', function () {
...
+            expect(
+              await allocatorWithManager.slotCanBeResetByAccount(
+                roles.airnode.address,
+                slotIndex,
+                roles.randomPerson.address
+              )
+            ).to.equal(false);
+          });
+        });
+        context('Setter of the slot is still a slot setter', function () {
```
acenolaza, line 498:
```suggestion
        context('Setter of the slot is a slot setter', function () {
```
Kind of confused by "still" 🤔

acenolaza (approved):

test/allocators/AllocatorWithAirnode.sol.js
bbenligiray: You're right, it should. Related to this, I'm planning to remove all public functions that expose roles (such as `deriveAdmin()` of AllocatorWithAirnode) because they can be derived off-chain/in the contract that need them. No one will make an external call if they can avoid it, I was being silly by cluttering the interface with these. Migrating these to `testUtils` was in fact preparation for that.

test/allocators/AllocatorWithManager.sol.js
bbenligiray: For them to have set a slot, they must have been a slot setter at one point. So I wanted to emphasize the distinction between someone who's not a slot setter and someone who was a slot setter at some point that set a slot and is no longer a slot setter.

test/allocators/AllocatorWithAirnode.sol.js
bbenligiray: I forgot to address this so created https://github.com/api3dao/airnode-protocol-v1/issues/86

PR #85, opened 2023-01-09
Refactor protocol/

bbenligiray:


Ashar2shahid (approved):
looks like only formatting changes (and a little bit of refactor in tests)

bbenligiray:
My man it took my whole day :skull:

test/test-utils.js
```diff
@@ -1,5 +1,14 @@
 const { ethers } = require('hardhat');
 
+const PROTOCOL_IDS = {
```
acenolaza, line 3: Maybe we need to move this object from [airnode repo ](https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/src/index.ts#L29-L36)to airnode-protocol-v1 repo and import it anywhere were it is needed.

Issue #53, opened 2022-11-26
Review docstrings


bbenligiray:
There have been some significant changes to how some of the contracts work. I tried to update the docstrings along with that, but I need to review it as a whole once the changes are all implemented.

Issue #40, opened 2022-11-24
Refactor tests


bbenligiray:
Currently tests are unrolled for better readability, but the overall size of the codebase caused extreme code duplication and in some test cases are not explored thoroughly to avoid further verbosity. The tests need to be refactored to address this issue.

bbenligiray:
As a part of this, add tests where the deviation reference is not zero

bbenligiray:
As a part of this, test the meta-tx cases for all contracts that support it

PR #88, opened 2023-01-13
Refactor dapis

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/53 https://github.com/api3dao/airnode-protocol-v1/issues/40 https://github.com/api3dao/airnode-protocol-v1/issues/39

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -17,14 +17,16 @@ import "./proxies/interfaces/IOevProxy.sol";
 /// asset price data feed. Beacons can also be seen as one-Airnode data feeds
 /// that can be used individually or combined to build Beacon sets. dAPIs are
 /// an abstraction layer over Beacons and Beacon sets.
+/// In addition, this contract allows winners of OEV auctions to pay their bids
+/// to update the specific data feed.
 /// @dev DapiServer is a PSP requester contract. Unlike RRP, which is
 /// implemented as a central contract, PSP implementation is built into the
 /// requester for optimization. Accordingly, the checks that are not required
 /// are omitted. Some examples:
 /// - While executing a PSP Beacon update, the condition is not verified
-/// because Beacon updates where the condition returns `false` (i.e., the
+/// because Beacon updates where the condition returns `false` (e.g., the
 /// on-chain value is already close to the actual value) are not harmful, and
-/// are even desirable.
+/// are even desirable ("any update is a good update").
 /// - PSP Beacon set update subscription IDs are not verified, as the
 /// Airnode/relayer cannot be made to "misreport a Beacon set update" by
 /// spoofing a subscription ID.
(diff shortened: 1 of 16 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -365,6 +378,10 @@ contract DapiServer is
                 this.fulfillPspBeaconUpdate.selector
             )
         );
+        require(
```
bbenligiray, line 381: The only functional change in this PR

contracts/dapis/DapiServer.sol
```diff
@@ -75,13 +77,17 @@ contract DapiServer is
     mapping(bytes32 => DataFeed) public override dataFeeds;
 
     /// @notice Data feed with ID specific to the OEV proxy
+    /// @dev This implies that an update as a result of an OEV auction only
+    /// affects contracts that read through the respective proxy that the
+    /// auction was being held for
```
acenolaza, line 82: I'm just checking if I understood this correctly. So this is saying that the update that resulted from an OEV auction being won will only change the data feed value that oev readers read. Not the value of the data feed that other readers read. (because oev readers read from this mapping and others like byog read directly from dataFeeds mapping)

contracts/dapis/DapiServer.sol
acenolaza: Kind of related to this. How does one un-registers a subscription? Would that even be allowed? Should we have a `resetBeaconSubscription()` function that resets the entry in the `subscriptionIdToHash` mapping?

contracts/dapis/DapiServer.sol
```diff
@@ -667,6 +689,10 @@ contract DapiServer is
     /// Airnodes without requiring a request or subscription. The Beacons for
     /// which the fulfillment data and signature is omitted will be read from
     /// storage.
+    /// @dev This signed data here is specific to this contract, which is to be
+    /// used when the signer does not want to provide the more general
+    /// signature. EIP712 may feel relevant here, but we avoided it for the
```
acenolaza, line 694: Probably dumb question but which is "the more general signature"?

contracts/dapis/DapiServer.sol
bbenligiray: The registration only has the DapiServer know what Beacon ID a subscription ID maps to. There is no downside to having it remain registered after the subscription stops being used. In contrast, unregistering an active subscription will prevent it from updating the Beacon from then on so I'm not sure who can be feasible allowed to do that, probably `airnode`, maybe `sponsor` but even that is iffy imo.

contracts/dapis/DapiServer.sol
bbenligiray: > So this is saying that the update that resulted from an OEV auction being won will only change the data feed value that oev readers read

So if project X is using ETH/USD with OEV, they have a special proxy that they read from. We hold an auction specific to that proxy address, and the related OEV updates only affect this proxy (or readers that pretend to be this proxy). Project Y using ETH/USD with OEV is not affected.

> others like byog read directly from dataFeeds mapping

I assume you mean self-funded data feeds and not byog the data provider. Self-funded data feeds will also be read through proxies, which optionally support OEV.

contracts/dapis/DapiServer.sol
bbenligiray: I mean the one from `updateDataFeedWithSignedData()`

PR #96, opened 2023-02-17
Remove signed Beacon set update

bbenligiray:
Addressing API3-01

- Removes domain-signed data updates altogether. I'll reintroduce it with the future PR that will rework OEV updates.
- Removes ability to update Beacon sets with signed data (and renames `updateDataFeedWithSignedData()` to be `updateBeaconWithSignedData()`)
- No longer requires the signed data to be encoded, which should reduce gas costs


acenolaza (approved):
LGTM 👍🏻

Ashar2shahid (approved):
looks good

PR #97, opened 2023-02-18
Median Beacon set timestamp

bbenligiray:
Addressing API3-02

- Uses median instead of mean to calculate Beacon set timestamps
- Allows `updateBeaconSetWithBeacons()` even if it doesn't update the Beacon set timestamp, as long as it updates the Beacon set value. This is because updating individual Beacons will not necessarily affect the Beacon set timestamp, but the resulting value will be fresher
- Small `updateBeaconWithSignedData()` refactor


acenolaza (approved):
LGTM 👍🏻 One minor thing tho is the comment in lines 881-882 which I don't think they still apply

Ashar2shahid (approved):
Looks good, I can see the auditors disliking the concept of updating the value without updating the timestamp because its not intuitive.

PR #98, opened 2023-02-18
Remove aggregation interface

bbenligiray:
Addresses API3-02

Makes `aggregateBeacons()` private. As a note, the response plan mentioned having `updateBeaconSetWithBeacons()` return the updated Beacon set to replace the `aggregateBeacons()` interface in function. However, this doesn't really work because `updateBeaconSetWithBeacons()` reverts if it can't update the Beacon set. The reader forcing Beacon set updates beforehand is not the intended user flow so I decided against reworking the interface just to enable this. This is still possible in a few ways:
- The reader `tryMulticall()`s a `updateBeaconSetWithBeacons()` and the read call
- The reader calls `updateBeaconSetWithBeacons()` in a try-catch statement
- The reader makes a low-level call to `updateBeaconSetWithBeacons()` and doesn't care if the call fails

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -876,32 +876,6 @@ contract DapiServer is
         require(timestamp > 0, "Data feed not initialized");
     }
 
-    /// @notice Called privately to aggregate the Beacons and return the result
-    /// @dev Tha aggregation of Beacons may have a different value than the
-    /// respective Beacon set, e.g., because the Beacon set has been updated
-    /// using signed data
-    /// @param beaconIds Beacon IDs
-    /// @return value Aggregation value
-    /// @return timestamp Aggregation timestamp
-    function aggregateBeacons(
-        bytes32[] memory beaconIds
-    ) private view returns (int224 value, uint32 timestamp) {
-        uint256 beaconCount = beaconIds.length;
-        require(beaconCount > 1, "Specified less than two Beacons");
-        int256[] memory values = new int256[](beaconCount);
-        int256[] memory timestamps = new int256[](beaconCount);
-        for (uint256 ind = 0; ind < beaconCount; ) {
-            DataFeed storage dataFeed = dataFeeds[beaconIds[ind]];
-            values[ind] = dataFeed.value;
-            timestamps[ind] = int256(uint256(dataFeed.timestamp));
-            unchecked {
-                ind++;
-            }
-        }
-        value = int224(median(values));
-        timestamp = uint32(uint256(median(timestamps)));
-    }
-
     /// @notice Derives the Beacon ID from the Airnode address and template ID
     /// @param airnode Airnode address
     /// @param templateId Template ID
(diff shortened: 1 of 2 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -996,6 +970,32 @@ contract DapiServer is
...
+    /// @notice Called privately to aggregate the Beacons and return the result
+    /// @dev Tha aggregation of Beacons may have a different value than the
+    /// respective Beacon set, e.g., because the Beacon set has been updated
+    /// using signed data
+    /// @param beaconIds Beacon IDs
+    /// @return value Aggregation value
+    /// @return timestamp Aggregation timestamp
+    function aggregateBeacons(
+        bytes32[] memory beaconIds
+    ) private view returns (int224 value, uint32 timestamp) {
```
acenolaza, line 982: I remember a discussion about this contract being sort of a base contract for other use cases of DapiServer in the future. Do you think this function might be useful in the contracts inheriting from this one? If so, should we make this function internal instead?

contracts/dapis/DapiServer.sol
```diff
@@ -996,6 +970,32 @@ contract DapiServer is
         }
     }
 
+    /// @notice Called privately to aggregate the Beacons and return the result
+    /// @dev Tha aggregation of Beacons may have a different value than the
+    /// respective Beacon set, e.g., because the Beacon set has been updated
+    /// using signed data
```
acenolaza, line 976: Can beacon sets still be updated using signed data?

acenolaza (approved):
LGTM 👍🏻 (just questions to double check things)

contracts/dapis/DapiServer.sol
acenolaza: nvm, found [this](https://github.com/api3dao/airnode-protocol-v1/pull/105/files#diff-47e178e1efdc14225ddb1d731a2e39c7b341830804023013586a1893265ef2b6L974-L976)

Ashar2shahid:
I'm looking at API3-02 and don't really see the reasoning behind making this private/internal. I can see this function being useful where dapps run a subgraph on top of DapiServer and call this function to get the most upto date price and then multicall the users call with updating the beaconSet. If no update is possible they would execute the user's call directly. This is just 1 use case

contracts/dapis/DapiServer.sol
bbenligiray: It feels like it's more likely that we will either not modify DapiServer or the modifications needed will be significant enough that we will not be able to reuse the current contracts without modification, and the things we do to make them more reusable will do nothing but reduce readability. I will see if these can be addressed in a final PR.

bbenligiray:
> the reasoning behind making this private/internal

This being publicly visible is not a security issue in the first place, but a usability issue. Reading the data feed with the Beacon set ID or calling `aggregateBeacons()` with the Beacon IDs that constitute the same Beacon set returns different values (as it should), even though they may appear to be equivalent (so the concern is that the users will fail to make the distinction).
- I'd prefer ticking off an audit issue to treating a made up, selfish usage scenario as a first class citizen
- The usage scenario is still supported in the ways listed in the description, though the user will have to update the Beacon set, which they should do anyway

bbenligiray:
(I now get what @Ashar2shahid means, this had nothing to do with the issue in the report)

PR #99, opened 2023-02-18
Do not invalidate old signatures

bbenligiray:
Addresses API3-03

- Since signed data can no longer be used to update Beacon sets without updating the respective Beacon, and updating the respective Beacon is only allowed if doing so increases its timestamp, we no longer need to invalidate old signatures. Omitting this validation is consistent with any old Beacon being usable to update a Beacon set.
- Removes the timestamp validation implementation in AirnodeRequester and tells the user to implement it themselves
- Increases the tolerance for timestamps from the future from 15 minutes to 1 hour. This was previously set at the low value of 15 minutes because using mean for timestamp aggregation enabled some attack scenarios. Now that we are using median for timestamp aggregation, a minority attacker group reporting large timestamps will not affect the aggregated timestamp, which means we can relax this tolerance to a more accommodating range.

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -1080,8 +1080,22 @@ contract DapiServer is
                 "Signature mismatch"
             );
             beaconValue = decodeFulfillmentData(data);
-            require(timestampIsValid(timestamp), "Timestamp not valid");
+            require(
+                timestamp != 0 && timestampIsValid(timestamp),
+                "Timestamp not valid"
+            );
             beaconTimestamp = uint32(timestamp);
         }
+    }
+
+    /// @notice Overriden to prevent a Beacon from being updated by an
+    /// unreasonably large timestamp, which would block all following updates
+    /// @dev The timestamp is allowed to be up to 1 hour later than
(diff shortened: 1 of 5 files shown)


acenolaza (approved):
LGTM 👍🏻

acenolaza (commented):
This also takes care of API3-06

contracts/dapis/DapiServer.sol
```diff
@@ -1080,8 +1080,22 @@ contract DapiServer is
...
+
+    /// @notice Overriden to prevent a Beacon from being updated by an
+    /// unreasonably large timestamp, which would block all following updates
+    /// @dev The timestamp is allowed to be up to 1 hour later than
+    /// `block.timestamp` to tolerate chain clock drift
+    /// @param timestamp Timestamp used in the signature
+    function timestampIsValid(
+        uint256 timestamp
+    ) internal view virtual override returns (bool) {
+        return timestamp < block.timestamp + 1 hours;
```
acenolaza, line 1099: Should we use `unchecked` for simple math operations like this one?

Ashar2shahid (approved):

PR #100, opened 2023-02-18
Remove template and subscription announcements

bbenligiray:
Addresses API3-04 (partially)

- Template and subscription announcements are removed

### DIFF
--- a/contracts/protocol/StorageUtils.sol
+++ b/contracts/protocol/StorageUtils.sol
@@ -3,16 +3,11 @@ pragma solidity 0.8.17;
 
 import "./interfaces/IStorageUtils.sol";
 
-/// @title Contract that can be used to announce template and subscription
-/// details through events and/or store them on-chain
+/// @title Contract that stores template and subscription details on chain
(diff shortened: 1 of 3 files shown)


acenolaza (approved):
LGTM 👍🏻 

Not checking on-chain that `msg.sender` is a sponsored requester was on purpose to make request consume less gas, right?

Ashar2shahid (approved):

bbenligiray:
That's only a side benefit. The reasoning is that sponsorship is something that is entirely up to the Airnode to honor, similar to authorizations. Then, checking this can be delegated to the node. This allows sponsorship schemes that are communicated off-chain. For example, consider a relayed RRP provider has an API that accepts signatures from a sponsor stating "I want to sponsor this requester address for up to 0.1 ETH for 2 days". Or a simpler example is the sponsorship being hardcoded in the Airnode config, similar to authorizations. So this is mostly for flexibility and enabling new usage scenarios, the requests costing less is a small side benefit (and spam susceptibility is a small disadvantage).

PR #101, opened 2023-02-18
Add notices about proxy constructor argument validation

bbenligiray:
Addresses API3-05

Since we deploy proxies as regular contracts for reduced external call overhead, we want to keep their bytecode small. Therefore, instead of adding checks in them, I added a notice to each of them to warn the user about this issue.


acenolaza (approved):
LGTM 👍🏻

Ashar2shahid (approved):
maybe change to `validated within ProxyFactory`

PR #102, opened 2023-02-19
Populate returndata of tryExternalMulticall() to EOA

bbenligiray:
Addresses API3-07

Populates the returndata of `tryExternalMulticall()` calls to EOAs in a way consistent with how `externalMulticall()` calls to EOAs revert

### DIFF
--- a/contracts/utils/ExternalMulticall.sol
+++ b/contracts/utils/ExternalMulticall.sol
@@ -75,6 +75,11 @@ contract ExternalMulticall is IExternalMulticall {
                 (successes[ind], returndata[ind]) = targets[ind].call(
                     data[ind]
                 );
+            } else {
+                returndata[ind] = abi.encodeWithSignature(
+                    "Error(string)",
+                    "Multicall target not contract"
+                );
             }
             unchecked {
                 ind++;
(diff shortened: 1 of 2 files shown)


contracts/utils/ExternalMulticall.sol
```diff
@@ -75,6 +75,11 @@ contract ExternalMulticall is IExternalMulticall {
                 (successes[ind], returndata[ind]) = targets[ind].call(
                     data[ind]
                 );
+            } else {
+                returndata[ind] = abi.encodeWithSignature(
+                    "Error(string)",
+                    "Multicall target not contract"
+                );
```
acenolaza, line 82: This code is really cool. We are simulating a revert by returning a revert error message but it didn't actually revert. Should we just return the string "Multicall target not contract" instead?

acenolaza (approved):

contracts/utils/ExternalMulticall.sol
Ashar2shahid: > "Multicall target not contract"

isn't that what we are doing ?

bbenligiray: If we `externalMulticall()` a reverting function, it reverts with the string. If we `externalMulticall()` a non-contract, it reverts with `"Multicall target not contract"`. If we `tryExternalMulticall()` a reverting function, it returns encoded revert data. For consistency, `tryExternalMulticall()`in a non-contract should return the respective encoded revert data, which is what I went with here.

bbenligiray: The only issue I have with it is that it would be better optimized to declare the encoded returndata as a constant, but I figured this is more readable and we don't need to worry about optimizing the gas cost for when EOAs are being called

PR #104, opened 2023-02-19
Make ExternalMulticall abstract and add guidance

bbenligiray:
Addresses API3-09

- Added some guidance to ExternalMulticall
- Made ExternalMulticall abstract so that it's a bit more difficult to deploy as it is and use it in a wrong way

### DIFF
--- a/contracts/utils/ExternalMulticall.sol
+++ b/contracts/utils/ExternalMulticall.sol
@@ -1,14 +1,25 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.17;
+pragma solidity ^0.8.0;
 
 import "./interfaces/IExternalMulticall.sol";
 
 /// @title Contract that enables calls to external contracts to be batched
-/// @notice This contract can be used to batch static calls or to interact with
-/// trusted contracts. Implements two ways of batching, one requires none of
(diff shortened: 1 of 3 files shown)


acenolaza (approved):
LGTM 👍🏻

contracts/utils/ExternalMulticall.sol
```diff
@@ -1,14 +1,25 @@
...
-contract ExternalMulticall is IExternalMulticall {
+/// @notice This contract can be used for two use-cases: (1) In its current
+/// state, it can be used to batch static calls to contracts that do not care
+/// about who the sender is, (2) after extending it, to interact with trusted
+/// contracts (see below for details). It implements two ways of batching, one
+/// requires none of the calls to revert and the other tolerates individual
+/// calls reverting.
+/// @dev As mentioned above, this contract can be used to interact with trusted
+/// contracts. Such interactions can leave this contract in a privileged
+/// position (e.g., with a non-zero balance of an ERC20 token), which can be
```
Ashar2shahid, line 15: not sure I quite understand the example

Ashar2shahid (approved):

contracts/utils/ExternalMulticall.sol
bbenligiray: Extended it

PR #105, opened 2023-02-20
Misc

bbenligiray:
Addresses API3-12, selectively

What I didn't address:
- The typecasting is indeed surprisingly expensive (~50 gas based on my tests). However, in addition to struct-packing reasons, using `uint32` in interfaces is useful in that it enforces additional type checks. There are only a few points where this optimization makes sense (because a lot of the time the timestamp will needed to be packed in a struct), and imo they are not worth creating special cases.
- We avoid custom errors because we may deploy on weird chains with custom block explorers that don't support them. We're not overly concerned about deployment costs (because our contracts are deployed once per chain) and optimizing the gas cost of the reverting branch is not a significant concern.
- We don't see a reason to limit role description lengths to 32 characters, and some of our role description lengths are already longer than that. Also, the fact that they are of string type is only a gas cost issue because we decided to provide the full role descriptions on-chain (because we will deploy these only once per chain so the additional gas cost is worth it). An AccessControlRegistry user implementation that is concerned about the additional gas cost should hardcode the description hashes instead.
- Dividing `require()` conditions that will result in the same error message being emitted reduces readability and the gas cost saving isn't worth it.

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -535,14 +535,13 @@ contract DapiServer is
     /// Even if this function returns `true`, the respective Subscription
     /// fulfillment will fail if will not update the Beacon set value or
     /// timestamp.
-    /// @param subscriptionId Subscription ID
     /// @param data Fulfillment data (array of Beacon IDs, i.e., `bytes32[]`
     /// encoded in contract ABI)
     /// @param conditionParameters Subscription condition parameters. This
     /// includes multiple ABI-encoded values, see `checkUpdateCondition()`.
     /// @return If the Beacon set update subscription should be fulfilled
     function conditionPspBeaconSetUpdate(
-        bytes32 subscriptionId, // solhint-disable-line no-unused-vars
+        bytes32,
         bytes calldata data,
         bytes calldata conditionParameters
     ) external view override returns (bool) {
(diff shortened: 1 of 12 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -572,23 +575,15 @@ contract DapiServer is
...
     /// matter for the purposes of a Beacon set update subscription.
-    /// @param subscriptionId Subscription ID
-    /// @param airnode Airnode address
-    /// @param relayer Relayer address
-    /// @param sponsor Sponsor address
-    /// @param timestamp Timestamp used in the signature
     /// @param data Fulfillment data (an `int256` encoded in contract ABI)
-    /// @param signature Subscription ID, timestamp, sponsor wallet address
-    /// (and fulfillment data if the relayer is not the Airnode) signed by the
-    /// Airnode wallet
```
acenolaza, line 583: WDYT of keeping the references like this:
```
    /// @param // subscriptionId Subscription ID
    /// @param // airnode Airnode address
    /// @param // relayer Relayer address
    /// @param // sponsor Sponsor address
    /// @param // timestamp Timestamp used in the signature
    /// @param data Fulfillment data (an `int256` encoded in contract ABI)
    /// @param // signature Subscription ID, timestamp, sponsor wallet address
    /// (and fulfillment data if the relayer is not the Airnode) signed by the
    /// Airnode wallet
```

contracts/dapis/DapiServer.sol
```diff
@@ -572,23 +575,15 @@ contract DapiServer is
...
-        address relayer, // solhint-disable-line no-unused-vars
-        address sponsor, // solhint-disable-line no-unused-vars
-        uint256 timestamp, // solhint-disable-line no-unused-vars
+        bytes32,
+        address,
+        address,
+        address,
+        uint256,
         bytes calldata data,
-        bytes calldata signature // solhint-disable-line no-unused-vars
```
acenolaza, line 591: And we can even keep the argument names like this for reference:

```
    function fulfillPspBeaconSetUpdate(
        bytes32 /*subscriptionId*/,
        address /*airnode*/,
        address /*relayer*/,
        address /*sponsor*/,
        uint256 /*timestamp*/,
        bytes calldata data,
        bytes calldata /*signature*/
```

acenolaza (approved):
LGTM 👍🏻

Ashar2shahid:
> The typecasting is indeed surprisingly expensive (~50 gas based on my tests). However, in addition to struct-packing reasons, using uint32 in interfaces is useful in that it enforces additional type checks. There are only a few points where this optimization makes sense (because a lot of the time the timestamp will needed to be packed in a struct), and imo they are not worth creating special cases.

Yeah I was looking into it a while ago, apparently casting to int256/uint256 is the cheapest (2 gas) as seen [here ](https://ethereum.stackexchange.com/questions/119947/what-are-the-gas-costs-for-type-casting-or-conversions)

> Dividing require() conditions that will result in the same error message being emitted reduces readability and the gas cost saving isn't worth it.

IDK, it makes the revert messages more specific(we don't have to reuse the same message in both the reverts) and saves gas too. might be worth it

Ashar2shahid (approved):

contracts/dapis/DapiServer.sol
bbenligiray: Strongly agree, did this

contracts/dapis/DapiServer.sol
bbenligiray: and this

bbenligiray:
If the conditions warranted different revert strings they would have been divided across requires in the first place

PR #107, opened 2023-02-21
Reworks OEV to aggregate off chain

bbenligiray:
Addresses the OEV portion of API3-01

The solution for signed updates in https://github.com/api3dao/airnode-protocol-v1/pull/96 doesn't work well for OEV updates because Beacons specific to OEV proxies are not live Beacons, i.e., the respective Airnode cannot update these on its own despite being censored by the OEV relay. OEV updates are multi-party process, which can be piggybacked on for a nicer solution.

Considering that an absolute majority of Beacon sets would be able to fully control the aggregation, it doesn't degrade the security guarantees to allow them to do the aggregation off-chain and write the aggregated value to the chain (except as seen above, the aggregation requirements are slightly relaxed). This resolves API3-01 by pushing the timestamp-related checks from the chain to the node. The main concern here is providing enough information to the honest nodes to be able to confidently come up with an aggregation. The questions to ask are
- Did the other nodes actually report these values? This can be verified through signatures. 
- Were the other nodes being censored (as in, "am I being fed outdated data even though these nodes would have provided more up to date if asked")? This is impossible to prove without a trusted channel, which is why outdated data has to be rejected. There are two things to consider here
  - We only need to check if the data point is outdated relative to the other data points. The contract will decide if the resulting aggregation improves the freshness of the on-chain data point (this is nice because we don't need to worry about chain time drift). 
  - Even when not all data is up to date, if an absolute majority is largely consistent on both value and timestamp, an aggregation from these can be achieved up to a confidence relative to how consistent they are. We can't simply require the absolute majority of the data points to be up to date and aggregate, as an attacking minority may then become over-represented in this absolute majority. Therefore, OEV Beacon set updates require the absolute majority to agree on both value and timestamp.

In practice, the OEV signing API of Airnode will receive signed data from all Beacons of the Beacon set, verify the signatures, calculate the median of values and timestamps, verify that an absolute majority falls within a confidence interval of this median (for example, in a 2 minutes and 1% window), then signs the aggregation along with all the OEV parameters.

### DIFF
--- a/contracts/dapis/DapiServer.sol
+++ b/contracts/dapis/DapiServer.sol
@@ -636,119 +636,117 @@ contract DapiServer is
 
     ///                     ~~~OEV~~~
 
-    /// @notice Updates a data feed that the OEV proxy reads using data signed
-    /// by the respective Airnodes for the specific bid. The Beacons for which
-    /// the fulfillment data and signature is omitted will be read from
-    /// storage.
-    /// @dev Even though data for Beacons are signed individually, the caller
-    /// is only allowed to use the signatures as a bundle. They cannot omit
-    /// individual signatures or mix-and-match among bundles.
+    /// @notice Updates a data feed that the OEV proxy reads using the
+    /// aggregation signed by the absolute majority of the respective Airnodes
+    /// for the specific bid
+    /// @dev For when the data feed being updated is a Beacon set, an absolute
+    /// majority of Airnodes must sign the resulting value and timestamp. While
+    /// doing so, Airnodes should refer to data signed by the respective
+    /// Airnodes. The absolute majority of this signed data should have similar
+    /// timestamps and values (e.g., within 2 minutes and 1% deviation).
+    /// Freshness should not be required, as OEV proxy data feeds cannot be
+    /// updated using older timestamps.
     /// @param oevProxy OEV proxy that reads the data feed
+    /// @param dataFeedId Data feed ID
     /// @param updateId Update ID
-    /// @param signatureCount Number of signatures in `signedData`
-    /// @param signedData Array of ABI-encoded Airnode address, template ID,
-    /// timestamp, fulfillment data and bid metadata that is signed by the
-    /// respective Airnode for the specific bid
+    /// @param timestamp Signature timestamp
+    /// @param data Update data (an `int256` encoded in contract ABI)
+    /// @param packedOevUpdateSignatures Packed OEV update signatures, which
+    /// include the Airnode address, template ID and these signed with the OEV
+    /// update hash
     function updateOevProxyDataFeedWithSignedData(
         address oevProxy,
+        bytes32 dataFeedId,
         bytes32 updateId,
-        uint256 signatureCount,
-        bytes[] calldata signedData
-    ) external payable override {
-        uint256 beaconCount = signedData.length;
-        bytes32 metadataHash = keccak256(
+        uint256 timestamp,
+        bytes calldata data,
+        bytes[] calldata packedOevUpdateSignatures
+    ) external payable override onlyValidTimestamp(timestamp) {
+        require(
+            timestamp > oevProxyToIdToDataFeed[oevProxy][dataFeedId].timestamp,
+            "Does not update timestamp"
+        );
+        bytes32 oevUpdateHash = keccak256(
             abi.encodePacked(
                 block.chainid,
                 address(this),
-                address(oevProxy),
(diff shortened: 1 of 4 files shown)


contracts/dapis/DapiServer.sol
```diff
@@ -636,119 +636,117 @@ contract DapiServer is
...
-                msg.value,
+                oevProxy,
+                dataFeedId,
                 updateId,
-                signatureCount,
-                beaconCount
+                timestamp,
+                data,
+                msg.sender,
+                msg.value
```
acenolaza, line 679: Since we are using `abi.encodePacked()`, shouldn't we put these 2 before the dynamically sized one (`data`)?

contracts/dapis/DapiServer.sol
```diff
@@ -636,119 +636,117 @@ contract DapiServer is
...
-                beaconId,
+                dataFeedId,
                 oevProxy,
                 updateId,
                 updatedValue,
                 updatedTimestamp
             );
         } else {
-            revert("Specified no Beacons");
+            revert("Did not specify any Beacons");
```
acenolaza, line 743: WDYT of moving this to a `require` at the beginning of the function to avoid wasting gas on accessing storage, encoding and decoding if `packedOevUpdateSignatures.length == 0`?

acenolaza (approved):
LGTM but I reviewed it to the best of my abilities focusing mostly on the Solidity code and not so much on the OEV workflow and business logic (which I'm still trying to get my head around)

contracts/dapis/DapiServer.sol
```diff
@@ -636,119 +636,117 @@ contract DapiServer is
...
         bytes32 updateId,
-        uint256 signatureCount,
-        bytes[] calldata signedData
-    ) external payable override {
-        uint256 beaconCount = signedData.length;
-        bytes32 metadataHash = keccak256(
+        uint256 timestamp,
+        bytes calldata data,
+        bytes[] calldata packedOevUpdateSignatures
+    ) external payable override onlyValidTimestamp(timestamp) {
```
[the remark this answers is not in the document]
bbenligiray: `onlyValidTimestamp()` is used because we don't want an update that has incorrectly used a wrong timestamp to permanently break the data feed. That is a bit risky because doing so will also block correct updates if there is chain time drift. `onlyValidTimestamp()` requires the timestamp to not be from the future more than 1 hour, assuming the chain time drift will be less than that.

The contract checking signature freshness is susceptible to blocking honest updates when there is chain time drift. When I said "we don't need to worry about chain time drift" I meant it in the sense that there won't be a freshness check. However, this doesn't make sense, because the problem with the contract checking signature freshness is that it checks this against the block time (which may drift), while the Airnode will be checking against the system time, so this is not a concern. I don't see why we would want to auction off outdated updates, so I will modify the comments to say that Airnodes will require the aggregation to be fresh to sign the respective OEV update.

contracts/dapis/DapiServer.sol
```diff
@@ -636,119 +636,117 @@ contract DapiServer is
...
+        bytes32 oevUpdateHash = keccak256(
             abi.encodePacked(
                 block.chainid,
                 address(this),
-                address(oevProxy),
-                msg.sender,
-                msg.value,
+                oevProxy,
+                dataFeedId,
                 updateId,
```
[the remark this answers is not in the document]
bbenligiray: This is an identifier generated randomly by the OEV Relay. If a searcher won two auctions we don't want them to be able to mix and match the signatures for the respective updates.

contracts/dapis/DapiServer.sol
bbenligiray: The ordering doesn't affect the size of the resulting packing while using `abi.encodePacked()` (it packs/doesn't pad). I think you're thinking about avoiding placing two consecutive dynamically sized variables in a packed encoding because it's ambiguous where one ends and the other one starts, which is not the case here because there is a single dynamically sized variable.

contracts/dapis/DapiServer.sol
bbenligiray: That would ever so slightly increase the gas cost of the happy paths. People should never call `updateOevProxyDataFeedWithSignedData()` with an empty `packedOevUpdateSignatures` array anyway so I sacrificed it. (I also expect `updateOevProxyDataFeedWithSignedData()` to be used with Beacon sets, which is why the first condition checks if a Beacon set is being updated.)

PR #108, opened 2023-02-24
Modularize dapiserver

bbenligiray:
- Breaks DapiServer down into modules.
- Creates an Api3ServerV1 out of these modules. Differently from the previous DapiServer, Api3ServerV1 doesn't allow Beacons to be updated using RRP or PSP
- Makes all data feed reading functions private and exposes them at Api3ServerV1. The reasoning here is that others may want to set up read-restricted data feed server contracts, now they will be able to reuse these components to do that.


README.md
```diff
@@ -60,14 +60,15 @@ users. Some mock contracts use custom errors, which is why they specify version
...
-  to create _Beacon sets_, which are asynchronously aggregated Beacons (we refer to both Beacons and Beacon sets as
-  "data feeds"). Finally, _dAPIs_ are names that are attached to Beacons or Beacon sets, which can be seen as
-  API3-managed data feeds. `DapiServer` uses RRP and PSP to update data feeds, which makes this contract a very good
-  example of how the Airnode protocol can be used. In addition, `DapiServer` allows data signed by the respective
-  Airnodes to update Beacons, which can be delivered outside of the protocol. Finally, a special version of these signed
-  data updates are used to capture
-  [oracle extractable value (OEV)](https://medium.com/api3/oracle-extractable-value-oev-13c1b6d53c5b).
+- `dapis/`: The `DataFeedServer` contract implements _Beacons_, which are single-Airnode data feeds. Beacons can be
+  combined to create _Beacon sets_, which are asynchronously aggregated Beacons (we refer to both Beacons and Beacon
+  sets as "data feeds"). _dAPIs_, implemented by `DapiServer`, are names that are attached to Beacons or Beacon sets,
```
acenolaza, line 65: 🤔 
```suggestion
  sets as "data feeds"). _dAPIs_, implemented by `DataFeedServer`, are names that are attached to Beacons or Beacon sets,
```

contracts/dapis/Api3ServerV1.sol
```diff
@@ -0,0 +1,85 @@
...
+import "./OevDapiServer.sol";
+import "./BeaconUpdatesWithSignedData.sol";
+import "./BeaconSetUpdatesWithPsp.sol";
+import "./interfaces/IApi3ServerV1.sol";
+
+/// @title First version of the contract that API3 uses to serve data feeds
+/// @notice Api3ServerV1 serves data feeds in the form of Beacons, Beacon sets,
+/// dAPIs, with optional OEV support for all of these.
+/// The base Beacons are only updateable using signed data, and the Beacon sets
+/// are updateable based on the Beacons, optionally using PSP. OEV proxy.
```
acenolaza, line 13:
```suggestion
/// are updateable based on the Beacons, optionally using PSP. OEV proxy,
```

acenolaza:
> * Creates an Api3ServerV1 out of these modules. Differently from the previous DapiServer, Api3ServerV1 doesn't allow Beacons to be updated using RRP or PSP

Is this why `updateBeaconSetWithBeacons()` function is defined in DataFeedServer.sol instead of having a separate contract like for the other update methods? (If we put that function in it's own contract then `BeaconSetUpdatesWithPsp` contract would then need to inherit this new contract)

acenolaza (approved):
I really like how everything was divided with the added benefit of being able to set different logic or checks for reading data feeds 👏🏻

bbenligiray:
> Is this a side effect of why updateBeaconSetWithBeacons() function is defined in DataFeedServer.sol instead of having a separate contract like for the other update methods? (If we put that function in it's own contract then BeaconSetUpdatesWithPsp contract would then need to inherit this new contract)

I don't think it would be a problem to implement `updateBeaconSetWithBeacons()` in a BeaconSetUpdatesWithBeacons contract and having BeaconSetUpdatesWithPsp inherit that. The main reason I didn't separate it is that I couldn't think of a use-case where you wouldn't want `updateBeaconSetWithBeacons()`. Maybe you wouldn't need it if you wouldn't be using Beacon sets, but then it's a nicer solution to implement a BeaconServer contract and having DataFeedServer inherit that, where `updateBeaconSetWithBeacons()` is still implemented in DataFeedServer. I didn't do that because I didn't want to go overboard with the number of contracts (though only using Beacons is probably a very legitimate use-case).

> BTW who is expected to use the functions in BeaconSetUpdatesWithPsp? I'm assuming that Airseeker will need to call updateBeaconWithSignedData and updateBeaconSetWithBeacons.

PSP is very costly for updating Beacons (compared to signed data) because subscriptions need to be registered, so the added flexibility is not worth it in most cases. However, PSP has no similar downside for updating Beacon sets. Also, unlike Beacon updates with PSP, Beacon set updates with PSP has no ill effects on OEV or API monetization (in that it doesn't deliver API-signed data anyway). So I kept it in for the lack of a reason to remove it, but there is no strong reason to keep it in either.

acenolaza:
> The main reason I didn't separate it is that I couldn't think of a use-case where you wouldn't want `updateBeaconSetWithBeacons()`

I guess one thing to keep in mind is that even when the `beaconSet` is updated via PSP the event being emitted would be `UpdatedBeaconSetWithBeacons`

PR #109, opened 2023-03-02
Extract BeaconServer from DataFeedServer

bbenligiray:


bbenligiray:
Closed because it's not clear if this is needed

Issue #93, opened 2023-01-15
Add `.code` to Multicall contracts


bbenligiray:
We sometimes need to check if multiple contracts have been deployed deterministically, being able to use multicall for that would be convenient

PR #111, opened 2023-03-02
Adds containsBytecode()

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/93

I figured I'd name the function to be less assertive than `isContract()`

### DIFF
--- a/contracts/utils/ExtendedSelfMulticall.sol
+++ b/contracts/utils/ExtendedSelfMulticall.sol
@@ -24,6 +24,18 @@ contract ExtendedSelfMulticall is SelfMulticall, IExtendedSelfMulticall {
         return account.balance;
     }
 
+    /// @notice Returns if the account contains bytecode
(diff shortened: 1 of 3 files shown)


acenolaza (approved):
LGTM 👍🏻

acenolaza:
Not very important but `OwnableCallForwarder` is already importing Address.sol so may be we could do `Address.isContract() `inside the `containsBytecode()` instead.

bbenligiray:
`ExtendedSelfMulticall` doesn't import `OwnableCallForwarder` though

acenolaza:
> `ExtendedSelfMulticall` doesn't import `OwnableCallForwarder` though

Right, I meant we could also import Address.sol like `OwnableCallForwarder` does. Sorry for my explaining it poorly 😁

bbenligiray:
Eh, I like the multicall contracts not having OpenZeppelin dependencies

PR #112, opened 2023-03-03
Remove BeaconSetUpdatesWithPsp from Api3ServerV1

bbenligiray:
See https://api3workspace.slack.com/archives/C027Y2FAV0S/p1677837115845199?thread_ts=1677834791.641009&cid=C027Y2FAV0S

I'd say BeaconSetUpdatesWithPsp and thus DataFeedUpdatesWithPsp are subject to change with the new heartbeat related discussion. Beacon sets can be updated with PSP even if Api3ServerV1 doesn't inherit BeaconSetUpdatesWithPsp, so we're not using anything.


.github/workflows/test.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v2
         with:
-          node-version: '14'
+          node-version: '18'
```
acenolaza, line 12: Just double checking...did you actually mean to commit this change here?

acenolaza (approved):
LGTM 👍🏻

.github/workflows/test.yml
bbenligiray: Yes, the workflow was failing without it https://github.com/api3dao/airnode-protocol-v1/actions/runs/4325151090/jobs/7550927220

PR #119, opened 2023-03-15
Rename DapiServer as Api3ServerV1 in proxies and factory

bbenligiray:
I updated the names on @Ashar2shahid's request. I think the api3ServerV1 name (rather than something more general) makes sense in that if one ever uses a custom data feed server implementation it will most likely not use proxies, so this proxy factory is in practice Api3ServerV1 specific.

### DIFF
--- a/contracts/dapis/proxies/DapiProxy.sol
+++ b/contracts/dapis/proxies/DapiProxy.sol
@@ -5,7 +5,7 @@ import "./interfaces/IDapiProxy.sol";
 import "../interfaces/IApi3ServerV1.sol";
 
 /// @title An immutable proxy contract that is used to read a specific dAPI of
-/// a specific DapiServer contract
+/// a specific Api3ServerV1 contract
 /// @notice In an effort to reduce the bytecode of this contract, its
 /// constructor arguments are validated by ProxyFactory, rather than
 /// internally. If you intend to deploy this contract without using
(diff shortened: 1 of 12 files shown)


acenolaza:
> so this proxy factory is in practice Api3ServerV1 specific.

but wouldn't this mean that we should also rename the proxy contracts? meaning contracts/dapis/proxies/DapiProxy.sol would then need to be contracts/dapis/proxies/Api3ServerV1Proxy.sol and contracts/dapis/proxies/DapiProxyWithOev.sol -> contracts/dapis/proxies/Api3ServerV1ProxyWithOev.sol, etc

Ashar2shahid (approved):
maybe the proxies can sit under `/proxies/api3ServerV1/*` in case we have proxies in the future for different version of DapiServer

bbenligiray:
> but wouldn't this mean that we should also rename the proxy contracts? meaning contracts/dapis/proxies/DapiProxy.sol would then need to be contracts/dapis/proxies/Api3ServerV1Proxy.sol and contracts/dapis/proxies/DapiProxyWithOev.sol -> contracts/dapis/proxies/Api3ServerV1ProxyWithOev.sol, etc

DapiServer -> DapiProxy
Api3ServerV1 -> Api3Proxy (?)

The "Dapi" in the proxy name refers to the type of data feed it's reading, not the name of the server contract. That being said, I can move all Api3ServerV1 related contracts (including the proxies and the factory) under `contracts/api3ServerV1/` if that helps.

Ashar2shahid:
I think just moving the proxies under `/proxies/api3ServerV1` would be good enough imo

PR #120, opened 2023-03-15
Add proxy address prediction functions

bbenligiray:
I figured I'd add this since https://github.com/api3dao/airnode-protocol-v1/pull/119 will require ProxyFactory to be redeployed anyway.

We will have `@api3/contracts` to export a function that predicts what a proxy address will be without deploying it. This address can also be obtained by performing a static call to one of the deployment functions. However, neither of these methods are available if it's a contract that needs to predict the address of a proxy. Instead, the user will have to implement the proxy address prediction function in their contract, and will likely make a mistake while doing (for example I forgot to hash the dAPI name while implementing this and only realized that when the tests reverted, another mistake I foresee is omitting the validations). So these are convenience functions, which also serve a security function.

### DIFF
--- a/contracts/dapis/proxies/ProxyFactory.sol
+++ b/contracts/dapis/proxies/ProxyFactory.sol
@@ -6,6 +6,7 @@ import "./DapiProxy.sol";
 import "./DataFeedProxyWithOev.sol";
 import "./DapiProxyWithOev.sol";
 import "./interfaces/IProxyFactory.sol";
+import "@openzeppelin/contracts/utils/Create2.sol";
 
 /// @title Contract factory that deterministically deploys proxies that read
 /// data feeds (Beacons or Beacon sets) or dAPIs, along with optional OEV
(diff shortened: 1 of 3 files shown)


acenolaza (approved):
I really like having these convenience functions and computing the address off-chain using ethers is not a trivial task so these can also be called off-chain without any associated cost 💪🏻

contracts/dapis/proxies/ProxyFactory.sol
```diff
@@ -111,4 +112,97 @@ contract ProxyFactory is IProxyFactory {
             metadata
         );
     }
+
+    /// @notice Computes the address of the data feed proxy
+    /// @param dataFeedId Data feed ID
+    /// @param metadata Metadata associated with the proxy
```
Ashar2shahid, line 118: maybe we can be a little more specific what metadata is in the @dev section

bbenligiray: I think we can't be more specific than "metadata" because this can be used for all kinds of stuff ranging from coverage to OEV to functionality that third parties implement

PR #121, opened 2023-03-16
Move files around

bbenligiray:


acenolaza:
It's kind of weird that `BeaconUpdatesWithSignedData` is not under the `/extensions` directory but I guess this is because it is built-in  the `Api3ServerV1` contract 🤔

acenolaza (approved):
Would make sense to move the *DapiServer*.sol and *DataFeedServer*.sol files to their own directories inside api3ServerV1 or would it just make code harder to follow?

PR #128, opened 2023-04-13
Add OrderPayable

bbenligiray:
Check this prototype out for managed feed payments

The backend has two options, it can issue signatures for orders that can be paid for by anyone, or it can issue signatures for invoices that are required to be paid by a specific account. My only concern about this is that it may confuse the user that will interact with the contract directly (because for example they will do the payment through a multisig/DAO proposal).


acenolaza:
On a first pass this looks good to me but I'd like to ask about the use case for limiting who pays the order. I don't seem to have enough context on when the `payInvoice()` could be called and if it is really necessary. 

Another doubt that I have is about `orderSigner`. Is this an EOA that the backend will control? I don't see this being a multisig and my concern has to do with keeping that mnemonic safe.

bbenligiray:
> the use case for limiting who pays the order

We want to reject payments from certain accounts (for example the OFAC-sanctioned ones). If we allow anyone to pay for an order and such an account ends up paying (1) it creates manual work for us to handle that (2) the order will be paid for and the actual user will be blocked from paying. It's not clear if this is an actual concern in practice because the sanctioned account will be losing a lot of money by doing this, so what I figured is we could have both in the contract, only use `payForOrder()` for now, switch to `payInvoice()` if the above scenario becomes a common occurrence. Another alternative is to have the contract only include `payForOrder()` and deploy a new contract that only has `payInvoice()` if that is ever needed.

> Is this an EOA that the backend will control? I don't see this being a multisig and my concern has to do with keeping that mnemonic safe.

`orderSigner` is an EOA. If it gets compromised, it can sign underpriced orders and have users pay for them. The payments here would still only be accessible by the multisig so we can simply revoke the role and refund the users.

acenolaza:
> We want to reject payments from certain accounts

I wonder if taking over the payment by using meta-txs would be better for this use case 🤔

acenolaza:
> If we generate an invoice for a non-blacklisted account and a blacklisted account pays for that using a meta-tx that doesn't achieve much

I was actually thinking about these steps:
1. No blacklist check when UI generates the order
2. User submits a meta-tx for a transfer
3. We ran the blacklist check 
4. Debit user's wallet if check passes

I still need to research if this is even possible 😄 and we would need to make txs on each chain

bbenligiray:
Right, two options then:
- They sign an actual transaction and send it to us, we execute it if we like the address it's from. Not nice because the signature doesn't expire, so they will have to send a transaction with the same nonce to make the signature invalid. This also means that if they send a transaction before we execute the transaction the signature will become invalid.
- They send funds to a contract, then sign a meta-tx for us to be able to withdraw funds from that contract on their behalf. This doesn't make much sense because the user may as well have sent the funds to OrderPayable and we would withdraw it from there.

Another issue with both is that the user needs an EOA

contracts/utils/OrderPayable.sol
```diff
@@ -0,0 +1,167 @@
...
+            _deriveAdminRole(manager),
+            WITHDRAWER_ROLE_DESCRIPTION
+        );
+    }
+
+    function payForOrder(
+        bytes32 orderId,
+        uint256 expirationTimestamp,
+        address orderSigner,
+        bytes calldata signature
```
bbenligiray, line 80: If we encode this into a single `bytes` it makes the multisig user flow simpler

contracts/utils/OrderPayable.sol
```diff
@@ -0,0 +1,167 @@
...
+        );
+        orderIdToPaymentStatus[orderId] = true;
+        emit PaidInvoice(
+            orderId,
+            expirationTimestamp,
+            orderSigner,
+            msg.sender,
+            msg.value
+        );
+    }
```
bbenligiray, line 147: @RiVal-cz wants to axe this for the initial version

bbenligiray: Removed [here](https://github.com/api3dao/airnode-protocol-v1/pull/128/commits/ea1a4f3c59f68acc5155d7c5d7c721cf48bf9d6c)

contracts/utils/OrderPayable.sol
bbenligiray: Done [here](https://github.com/api3dao/airnode-protocol-v1/pull/128/commits/bbe39cf7a77a19953e110c3bf492b871382d634a)

bbenligiray:
Finalized the implementation based on discussions with @RiVal-cz and @umersin61

contracts/utils/OrderPayable.sol
```diff
@@ -0,0 +1,142 @@
...
+pragma solidity 0.8.17;
+
+import "../access-control-registry/AccessControlRegistryAdminnedWithManager.sol";
+import "./interfaces/IOrderPayable.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+
+/// @title Contract used to pay for orders denoted in the native currency
+/// @notice OrderPayable is managed by an account that designates order signers
+/// and withdrawers. Only orders for which a signature is issued for by an
+/// order signed can be paid for. Order signers have to be EOAs to be able to
```
acenolaza, line 11: 🤔 
```suggestion
/// order signer can be paid for. Order signers have to be EOAs to be able to
```

contracts/utils/interfaces/IOrderPayable.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IOrderPayable {
+    event PaidForOrder(
+        bytes32 indexed orderId,
+        uint256 expirationTimestamp,
+        address orderSigner,
+        uint256 amount,
+        address sender
```
acenolaza, line 10: I wonder if this should also be `indexed`? 🤔

contracts/utils/interfaces/IOrderPayable.sol
```diff
@@ -0,0 +1,38 @@
...
+    );
+
+    event Withdrew(address recipient, uint256 amount);
```
acenolaza, line 13: Same with recipient 🤔

contracts/utils/OrderPayable.sol
```diff
@@ -0,0 +1,142 @@
...
+        require(expirationTimestamp > block.timestamp, "Order expired");
+        require(
+            orderSigner == manager ||
+                IAccessControlRegistry(accessControlRegistry).hasRole(
+                    orderSignerRole,
+                    orderSigner
+                ),
+            "Invalid order signer"
+        );
+        require(msg.value > 0, "Payment amount zero");
```
acenolaza, line 96: It might be better to move this check at the beginning to save some gas by avoiding the `abi.decode`, etc

acenolaza (approved):
LTGM 👍🏻 Looking forward to see this in action on the test file 💪🏻

contracts/utils/interfaces/IOrderPayable.sol
bbenligiray: At least the intended user flow doesn't require it. We need to fetch all events and go through all `sender`s one by one to check them. Arguably even `orderId` doesn't need to be indexed.

contracts/utils/interfaces/IOrderPayable.sol
bbenligiray: Same here, I can't think of a user flow that requires us to get all withdrawals to a specific address. When we need it we'll need all of it.

contracts/utils/OrderPayable.sol
bbenligiray: I wanted to first decode the arguments, then validate the arguments in the order that they're used in the signature and not care about optimizing the unhappy path

contracts/utils/interfaces/IOrderPayable.sol
Ashar2shahid: I agree with it being indexed, could help with things like third party services that track payments/proxy deployments etc

contracts/utils/OrderPayable.sol
```diff
@@ -0,0 +1,142 @@
...
+        require(!orderIdToPaymentStatus[orderId], "Order already paid for");
+        require(
+            (
+                keccak256(
+                    abi.encodePacked(
+                        block.chainid,
+                        address(this),
+                        orderId,
+                        expirationTimestamp,
```
Ashar2shahid, line 105: shouldn't we add `msg.sender` to this signature. I know we will have a check in the backend but I don't think anyone should be able to execute orders if it's signed by the backend.

PR #129, opened 2023-04-15
Add Funder contract prototypes

bbenligiray:
I implemented some contract prototypes for funding sponsor wallets

FunderLite: This is a minimal implementation that is fairly gas-optimized
Funder + FunderFactory: This allows you to fund 100 separate sponsor wallets for each Nodary Beacon (which I think is an actual use-case)

I provided FunderLite in case you find Funder too heavy for dAPIs team use-case. However, considering `lowThreshold` and `highThreshold` will be fairly different (so `fund()` will be called fairly infrequently and it being heavy doesn't matter that much) and it's better to have one funding solution that we perfect, I think Funder can also be used for dAPI purposes.

Goes without saying but this contract can also be used to fund RRP sponsor wallets, for example for QRNG.

### DIFF
--- a/contracts/utils/Funder.sol
+++ b/contracts/utils/Funder.sol
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
+
+// I wanted to build this in a way that you can also use it to fund 100 sponsor
+// wallets that each power a single self-funded Beacon.
+// Use FunderFactory to deploy.
+contract Funder is Ownable {
+    // Making this immutable would be nice. However, I foresee that we will
+    // want to update this and having to transfer funds to do so will be very
+    // annoying/error-prone.
+    bytes32 public root;
+
+    // This contract is intended to be cloned so the constructor makes the
+    // implementation unusable
+    constructor() {
+        _transferOwnership(address(0));
+        root = bytes32(uint256(1));
+    }
+
+    // Anyone can send ETH to this contract, it can only be used in funding or
+    // withdrawn by the owner
+    receive() external payable {
+        // Maybe emit an event? If the user is withdrawing from an exchange for
+        // example and the exchange hardcodes the gas limit as 21,000, the
+        // event will have that transaction revert. That being said, maybe even
+        // the `receive()` itself causes enough overhead to have this use-case
+        // fail. I will experiment on this.
+    }
+
+    // `initialOwner` can be `address(0)` to make the root immutable and not
+    // allow funds to be withdrawn
+    function initialize(address initialOwner, bytes32 initialRoot) external {
+        require(root == bytes32(0), "Cannot initialize");
+        _transferOwnership(initialOwner);
+        root = initialRoot;
+        // Emit event
+    }
+
+    function setRoot(bytes32 _root) external onlyOwner {
+        require(_root != bytes32(0), "Root zero");
+        root = _root;
+        // Emit event
+    }
+
+    // I figure we may want to skim without disturbing its operation so I added
+    // `amount`
+    function withdraw(address recipient, uint256 amount) external onlyOwner {
+        require(recipient != address(0), "Recipient address zero");
+        require(amount != 0, "Amount zero");
+        require(address(this).balance >= amount, "Insufficient balance");
+        // Emit event
+        (bool success, ) = recipient.call{value: amount}("");
+        require(success, "Transfer unsuccessful");
+    }
+
+    // One can build a layer on top of this to reward accounts that call this.
+    // An even cooler thing to do would be to set `root` so that this function
+    // also funds the reward contract.
+    function fund(
+        bytes32[] calldata proof,
+        address recipient,
+        uint256 lowThreshold,
+        uint256 highThreshold
+    ) external {
+        // Still checking these in case the tree was populated with invalid
+        // values
+        require(recipient != address(0), "Recipient address zero");
+        require(
+            lowThreshold <= highThreshold,
+            "Low threshold higher than high"
+        );
+        require(highThreshold != 0, "High threshold zero");
+        // https://github.com/OpenZeppelin/merkle-tree#validating-a-proof-in-solidity
+        bytes32 leaf = keccak256(
+            bytes.concat(
+                keccak256(abi.encode(recipient, lowThreshold, highThreshold))
+            )
+        );
+        require(MerkleProof.verify(proof, root, leaf), "Invalid proof");
+        // https://en.wikipedia.org/wiki/Hysteresis#In_engineering
+        uint256 recipientBalance = recipient.balance;
+        require(recipientBalance <= lowThreshold, "Balance not low enough");
+        uint256 topUpAmount = highThreshold - recipientBalance <
+            address(this).balance
+            ? highThreshold - recipientBalance
+            : address(this).balance;
+        require(topUpAmount != 0, "Top up amount zero");
+        // Emit event
+        (bool success, ) = recipient.call{value: topUpAmount}("");
+        require(success, "Transfer unsuccessful");
+    }
+}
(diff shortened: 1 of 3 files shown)


contracts/utils/Funder.sol
```diff
@@ -0,0 +1,95 @@
...
+    // Making this immutable would be nice. However, I foresee that we will
+    // want to update this and having to transfer funds to do so will be very
+    // annoying/error-prone.
+    bytes32 public root;
+
+    // This contract is intended to be cloned so the constructor makes the
+    // implementation unusable
+    constructor() {
+        _transferOwnership(address(0));
+        root = bytes32(uint256(1));
```
bdrhn9, line 20: Was this something intended to initialize `root` with `1`? Probably I'm missing sth.

contracts/utils/Funder.sol
```diff
@@ -0,0 +1,95 @@
...
+            lowThreshold <= highThreshold,
+            "Low threshold higher than high"
+        );
+        require(highThreshold != 0, "High threshold zero");
+        // https://github.com/OpenZeppelin/merkle-tree#validating-a-proof-in-solidity
+        bytes32 leaf = keccak256(
+            bytes.concat(
+                keccak256(abi.encode(recipient, lowThreshold, highThreshold))
+            )
+        );
```
acenolaza, line 81: Just being curious, is there a case where we might want to have multiple leafs for the same `recipient` but with different `lowThreshold` and `highThreshold`?

contracts/utils/Funder.sol
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/cryptography/MerkleProof.sol";
+
+// I wanted to build this in a way that you can also use it to fund 100 sponsor
+// wallets that each power a single self-funded Beacon.
+// Use FunderFactory to deploy.
+contract Funder is Ownable {
```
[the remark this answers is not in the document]
acenolaza: I'm guessing it will be switched from `Ownable` to `OwnableCallForwarder` as well, right?

contracts/utils/Funder.sol
acenolaza: I was wondering the same thing, it feels like `root` should be initialized with `bytes32(0)` or `0x00`

bbenligiray: It's so that the implementation can't be initialized

contracts/utils/Funder.sol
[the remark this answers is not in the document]
bbenligiray: SelfMulticall is for when one needs to fund multiple sponsor wallets with a single transaction. I don't see why OwnableCallForwarder would be needed.

contracts/utils/FunderLite.sol
```diff
@@ -0,0 +1,41 @@
...
+    receive() external payable {}
+
+    // No Merkle proofs. Still allows reward mechanism to be built on top.
+    function fund() external {
+        uint256 recipientBalance = recipient.balance;
+        require(recipientBalance <= lowThreshold, "Balance not low enough");
+        uint256 topUpAmount = highThreshold - recipientBalance <
+            address(this).balance
+            ? highThreshold - recipientBalance
+            : address(this).balance;
```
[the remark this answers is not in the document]
bbenligiray: `unchecked` is cheaper because it omits that `require` so I don't think so. Keeping `highThreshold - recipientBalance` and `address(this).balance` in variables would improve things though. I wanted to keep things readable for now.

contracts/utils/FunderFactory.sol
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    function computeFunderAddress(
+        address initialOwner,
+        bytes32 initialRoot
+    ) external view returns (address funder) {
+        funder = Clones.predictDeterministicAddress(
+            funderImplementation,
+            salt(initialOwner, initialRoot)
+        );
```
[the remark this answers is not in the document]
bbenligiray: Good point, probably not :expressionless:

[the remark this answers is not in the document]
bbenligiray: Perhaps also considering [this](https://github.com/api3dao/airnode-protocol-v1/blob/self-funded/contracts/utils/FunderFactory.sol#L14-L19), cloning should be undeterministic

contracts/utils/Funder.sol
alikonuk1: To make it clearer, we're using cloning method to deploy the Funder.sol contract. The constructor does not execute during deployment, and the contract is initialized directly. So `root == bytes32(0)` will be validated when initializing. This makes the deployment more efficient and saves gas.

contracts/utils/Funder.sol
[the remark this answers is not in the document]
acenolaza: I guess I got confused about who the owner of this contract is going to be. I thought It would be the DAO or some multisig that might not interact with this contract directly.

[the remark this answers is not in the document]
bbenligiray: > I thought It would be the DAO or some multisig that might not interact with this contract directly.

That's correct. I still don't get what call you would have this contract forward though.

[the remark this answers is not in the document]
bbenligiray: Done in https://github.com/api3dao/airnode-protocol-v1/pull/129/commits/bd9d03f813c120f78097c5a228729825deb4c800

contracts/utils/FunderLite.sol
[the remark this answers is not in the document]
bbenligiray: Done in https://github.com/api3dao/airnode-protocol-v1/pull/129/commits/d88a0669a86e4afe34972582ade3164b53a9a58c

contracts/utils/FunderFactory.sol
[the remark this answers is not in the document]
bbenligiray: See https://github.com/api3dao/airnode-protocol-v1/pull/129#issuecomment-1515970605

contracts/utils/FunderV2.sol
```diff
@@ -0,0 +1,118 @@
...
+        require(root != bytes32(0), "Root zero");
+        funderDepository = payable(
+            new FunderDepositoryV2{salt: bytes32(0)}(owner, root)
+        );
+        ownerToRootToFunderDepositoryAddress[owner][root] = funderDepository;
+        // Emit event
+    }
+
+    function withdraw(
+        address owner,
```
acenolaza, line 29: This argument seems redundant since we are checking that is equal to msg.sender, right?

bbenligiray: Good point

alikonuk1 (approved):

contracts/utils/Funder.sol
Ashar2shahid: why not something like:
```
root = bytes32(bytes("CANNOT_INITIALIZE"));
```

bbenligiray: Would work just as well

contracts/utils/Funder.sol
```diff
@@ -0,0 +1,119 @@
...
+            owner
+        ][root];
+        uint256 amountNeededToTopUp;
+        unchecked {
+            amountNeededToTopUp = highThreshold - recipientBalance;
+        }
+        uint256 balance = funderDepository.balance;
+        uint256 amount = amountNeededToTopUp <= balance
+            ? amountNeededToTopUp
+            : balance;
```
alikonuk1, line 68: Since it will not get called for both path at once, we can save 5 gas by not initializing a new `balance` value and just calling `funderDepository.balance` on both path. 

```suggestion
        uint256 amount = amountNeededToTopUp <= funderDepository.balance
            ? amountNeededToTopUp
            : funderDepository.balance;
```

bbenligiray: I didn't test it but I suspect that this optimizes the case of `amountNeededToTopUp <= funderDepository.balance` at the expense of the `amountNeededToTopUp > funderDepository.balance` case. In the latter case, `funderDepository.balance` is used twice (first to check the condition and second to assign to `amount`), and assigning `funderDepository.balance` to `balance` and using that should be cheaper.
However, `amountNeededToTopUp <= funderDepository.balance` can be considered to be the happy path, so this is a viable optimization.

PR #130, opened 2023-04-21
Add tests for OrderPayable

alikonuk1:
Added tests that covers various cases and validates the expected behavior for functions in OrderPayable.sol, which will be added on this pr #128


test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+  describe('constructor', function () {
+    it('constructor', async function () {
+      const { roles, orderPayable } = await helpers.loadFixture(deploy);
+      expect(await orderPayable.manager()).to.equal(roles.manager.address);
+    });
+  });
+
+  describe('payForOrder', function () {
+    context('Order signer valid', function () {
+      context('Order Id not zero', function () {
```
bbenligiray, line 47: As a personal stylistic choice, I use regular English in `context()` strings. I'd prefer `Order signer is valid` (we don't need to shorten this as this is not a revert string) and `Order Id not zero` (this is not code, we don't need to use camel case).

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+        context('Order not expired', function () {
+          context('Payment amount not zero', function () {
+            context('Order didnt paid before', function () {
```
bbenligiray, line 50: Missing apostrophe, broken grammar. Contract code is formal writing so you shouldn't use contractions anyway (so `did not` over `didn't`).

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+              context('Signiture match', function () {
```
bbenligiray, line 51: Typo

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+
+  describe('constructor', function () {
+    it('constructor', async function () {
+      const { roles, orderPayable } = await helpers.loadFixture(deploy);
+      expect(await orderPayable.manager()).to.equal(roles.manager.address);
+    });
+  });
+
+  describe('payForOrder', function () {
+    context('Order signer valid', function () {
```
bbenligiray, line 46: Context is `Order signer valid` but there is no alternative context, so this is redundant

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+  }
+
+  describe('constructor', function () {
+    it('constructor', async function () {
+      const { roles, orderPayable } = await helpers.loadFixture(deploy);
+      expect(await orderPayable.manager()).to.equal(roles.manager.address);
+    });
+  });
+
+  describe('payForOrder', function () {
```
bbenligiray, line 45: The conditions should be tested in the order that they are validated. For example, for `payForOrder()`
- Order ID needs to be non-zero
- Order needs to not have expired
- Signer needs to be the manager or an order signer (these need to be tested separately)
- Payment amount needs to be non-zero
- Order needs to not have been paid for
- Signature has to match

Your contexts should follow the same order to make sure that you're not missing any branches

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+                  );
+
+                  const hash = ethers.utils.arrayify(hashedMessage);
+
+                  const signature = await roles.manager.signMessage(ethers.utils.arrayify(hash));
+
+                  const encodedData = ethers.utils.defaultAbiCoder.encode(
+                    ['bytes32', 'uint256', 'address', 'bytes'],
+                    [orderId, expirationTimestamp, orderSignerAddress, signature]
+                  );
```
bbenligiray, line 76: Duplicated too many times, extract this as a function

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+                  const { roles, orderPayable } = await deploy();
+
+                  const orderId = testUtils.generateRandomBytes32();
+                  const timestamp = await helpers.time.latest();
+                  const expirationTimestamp = timestamp + 60;
+                  const paymentAmount = ethers.utils.parseEther('1');
+                  const orderSignerAddress = roles.manager.address;
+
+                  const network = await orderPayable.provider.getNetwork();
+                  const chainId = network.chainId;
```
bbenligiray, line 62: Can be a single line since `network` isn't used again

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+            expect(await ethers.provider.getBalance(orderPayable.address)).to.equal(ethers.utils.parseEther('0'));
+            expect(await orderPayable.orderIdToPaymentStatus(orderId)).to.equal(false);
+          });
+        });
+      });
+      context('Order Id zero', function () {
+        it('Target function reverts', async function () {
+          const { roles, orderPayable } = await deploy();
+
+          const orderId = ethers.utils.hexZeroPad(
```
bbenligiray, line 195: https://docs.ethers.org/v5/api/utils/constants/#constants-HashZero

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+          ).to.be.revertedWith('Order ID zero');
+          expect(await ethers.provider.getBalance(orderPayable.address)).to.equal(ethers.utils.parseEther('0'));
+          expect(await orderPayable.orderIdToPaymentStatus(orderId)).to.equal(false);
+        });
+      });
+    });
+  });
+
+  describe('withdraw', function () {
+    context('Caller is manager or withdrawer', function () {
```
bbenligiray, line 232: Test separately

test/utils/OrderPayable.sol.js
```diff
@@ -0,0 +1,319 @@
...
+            orderPayable.connect(roles.manager).payForOrder(encodedData, { value: paymentAmount })
+          ).to.be.revertedWith('Order ID zero');
+          expect(await ethers.provider.getBalance(orderPayable.address)).to.equal(ethers.utils.parseEther('0'));
+          expect(await orderPayable.orderIdToPaymentStatus(orderId)).to.equal(false);
+        });
+      });
+    });
+  });
+
+  describe('withdraw', function () {
```
bbenligiray, line 231: You should test the case where the value transfer reverts. You can do that by setting the `recipient` to be a contract that doesn't have a `receive()` function (for example AccessControlRegistry).

bbenligiray (changes requested):
I ran `test:coverage` and saw that OrderPayable is not 100% covered. Anything less than 100% is not acceptable for contracts, and even 100% coverage by itself does not guarantee good tests.

alikonuk1:
> I ran `test:coverage` and saw that OrderPayable is not 100% covered. Anything less than 100% is not acceptable for contracts, and even 100% coverage by itself does not guarantee good tests.

I have updated the tests for OrderPayable contract, now it have 100% coverage. I would love to know if there is still an area for improvement.

bbenligiray (approved):
I applied some changes [here](https://github.com/api3dao/airnode-protocol-v1/pull/134)

PR #131, opened 2023-04-24
Add value forwarding support to ExternalMulticall

### LINKED ISSUE #106: Allow ExternalMulticall to forward value
One of the uses of ExternalMulticall is to provide a base for searcher contracts. Searchers may need to transfer a `msg.value` along with function calls while interacting with the contracts they are extracting OEV from, and they definitely need to do so while executing the updates to pay the bid amount. ExternalMulticall needs to be able to forward values to enable this.

alikonuk1:
This PR closes #106 by adding support for forwarding values in the `ExternalMulticall` contract.

- Updated the `externalMulticall` and  `tryExternalMulticall` functions to accept an additional `uint256[] calldata values` parameter, which represents the value to be sent with each function call.
- Added a loop to sum the values passed in the `values` array and added a check to see if the `msg.value` is enough to send every individual value.
- Updated the `call()` method with the `{value: values[ind]}` option to forward the specified value with each function call.

These changes enable the `ExternalMulticall` contract to forward the value along with function calls, allowing searchers to transfer a `msg.value` while interacting with the contracts they are extracting OEV from and to pay the bid amount when executing updates.

I have also made changes to other related contracts as necessary. In the test file `ExternalMulticall.sol.js`, I added the `values` parameter to the function arguments and included the `msg.value` in the call. Proper testing for these changes will be added.


acenolaza (commented):
Implementation LGTM 🚀 

Just one thought, I wonder if it would be better to have created separate functions (i.e. `externalMulticallWithValue()`, etc) to avoid having to send an array full of zeros when no value is sent on any of the calldatas. This way we avoid iterating over the `values` array to sum all the zeros in it. BTW, is there a test to test when value is being set along side the calldatas?

alikonuk1:
Also, when calling `tryExternalMulticall`, if one of the forwarded calls not succeeds, value sent to the contract will not get forwarded and will stuck in the contract. 
Maybe we can add a refund mechanism at the end of `tryExternalMulticall` function that refunds the value equal to the balance of the contract at that point.

acenolaza:
> Also, when calling `tryExternalMulticall`, if one of the forwarded calls not succeeds, value sent to the contract will not get forwarded and will stuck in the contract. Maybe we can add a refund mechanism at the end of `tryExternalMulticall` function that refunds the value equal to the balance of the contract at that point.

I was supposed to mention something about funds getting stuck in the contract during the review but then forgot 🤦🏻 

Since we are checking that `msg.value` is equal **or greater** than what's required 
(`require(msg.value >= value, "Insufficient value");`) I thought that there might be a case when there is a leftover after a multicall tx. So I'm not so sure about sending the whole balance 🤔 This function might need to be invoked by an address with a privileged role or might need a signature generated by a specific wallet.

bbenligiray:
> I wonder if it would be better to have created separate functions

I like that, very similar to `functionCall()` vs. `functionCallWithValue()` from [Address.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Address.sol)

> Also, when calling tryExternalMulticall, if one of the forwarded calls not succeeds, value sent to the contract will not get forwarded and will stuck in the contract.

That's indeed very problematic. I'd say that the correct way to address that is through a sweep function that can withdraw funds from the contract, but this sweep function will now need to be access-controlled, and in the case that ExternalMulticall is inherited by a contract that may have a non-zero balance another use-case using the sweep function may result in unexpected behavior. Instead, I'd question if `tryExternalMulticallWithValue()` is needed at all. Maybe we just need `externalMulticall()`, `tryExternalMulticall()` and `externalMulticallWithValue()`. At least the searcher use-case would always want to use `externalMulticallWithValue()` over `tryExternalMulticallWithValue()`.

contracts/utils/ExternalMulticall.sol
```diff
@@ -24,13 +24,26 @@ abstract contract ExternalMulticall is IExternalMulticall {
...
+            "Parameter length mismatch"
+        );
+        uint256 value = 0;
+        for (uint256 i = 0; i < callCount; ) {
+            value += values[i];
+            unchecked {
+                ++i;
+            }
+        }
+        require(msg.value >= value, "Insufficient value");
```
bbenligiray, line 46: We should require an exact match here because the opposite is guaranteed to be a user error

contracts/utils/ExternalMulticall.sol
```diff
@@ -24,13 +24,26 @@ abstract contract ExternalMulticall is IExternalMulticall {
...
+            callCount == data.length && callCount == values.length,
+            "Parameter length mismatch"
+        );
+        uint256 value = 0;
+        for (uint256 i = 0; i < callCount; ) {
+            value += values[i];
+            unchecked {
+                ++i;
+            }
+        }
```
bbenligiray, line 45: This loop is wasteful, this should happen in the loop below. Something like
```
uint256 accumulatedValue = 0;
for (uint256 ind = 0; ind < callCount; ) {
    accumulatedValue += values[ind];
    require(msg.value >= accumulatedValue, "Insufficient value");
    require(
        targets[ind].code.length > 0,
        "Multicall target not contract"
    );
    bool success;
    // solhint-disable-next-line avoid-low-level-calls
    (success, returndata[ind]) = targets[ind].call{ value: values[ind] } (
        data[ind]
    );
    if (!success) {
        bytes memory returndataWithRevertData = returndata[ind];
        // Adapted from OpenZeppelin's Address.sol
        if (returndataWithRevertData.length > 0) {
            // solhint-disable-next-line no-inline-assembly
            assembly {
            let returndata_size:= mload(returndataWithRevertData)
            revert(
                add(32, returndataWithRevertData),
                returndata_size
            )
            }
        } else {
            revert("Multicall: No revert string");
        }
    }
    unchecked {
        ind++;
    }
}
require(msg.value == accumulatedValue, "Excess value");
```

bbenligiray:
On a separate note, the user of ExternalMulticall will in most cases use only one of the four options (with/without try and value). ExternalMulticall is currently declared as a `contract`, which means they will still have to deploy the bytecode of all four implementations. We should instead declare it as a `library` so the user needs to import and expose the functions they need. However, `library` functions can't have the `payable` modifier, so we can't do this for the WithValue functions. I'm not sure how to handle this other than keeping each function in a separate contract.

alikonuk1:
> Maybe we just need `externalMulticall()`, `tryExternalMulticall()` and `externalMulticallWithValue()`. At least the searcher use-case would always want to use `externalMulticallWithValue()` over `tryExternalMulticallWithValue()`.

Certainly, I agree. This approach seems to be more  practical. I have updated the implementation to include those three functions: `externalMulticall()`, `tryExternalMulticall()`, and `externalMulticallWithValue()`.

alikonuk1:
> We should instead declare it as a `library` so the user needs to import and expose the functions they need. However, `library` functions can't have the `payable` modifier, so we can't do this for the WithValue functions. I'm not sure how to handle this other than keeping each function in a separate contract.

Great point. We can use a separate `library` for non-payable functions ( `externalMulticall()` and `tryExternalMulticall()` ) and an `abstract contract` for payable function `externalMulticallWithValue()`. This way bytecode size can be reduced and the codebase can be more modular. Only con I can see here would be increased complexity maybe? for the user, as they would need to manage both libraries and abstract contracts. But this trade-off seems acceptable in order to achieve modularity and decreased bytecode size.

bbenligiray (changes requested):
Does not have 100% test coverage

bbenligiray:
I like the idea of `externalMulticallWithValue()` being in a standalone `abstract contract`, as it will always need to be inherited. `externalMulticall()` and `tryExternalMulticall()` work on their own/don't need to be inherited (we'll mostly need them for batch-reading through static calls) so they can remain the same. So I'd say leave `ExternalMulticall` as is (i.e., revert all changes to it), create a new `ExternalMulticallWithValue` as an `abstract contract` (and move the tests to a new file accordingly).

Siegrift (changes requested):
I've tried using this for OEV and realized that the contract changes `msg.sender` which causes the transaction to revert, because msg.sender is checked [here](https://github.com/api3dao/airnode-protocol-v1/blob/3553a52c483525caa3afd2d244c99da75ca10335/contracts/api3-server-v1/OevDataFeedServer.sol#L66).

See: https://github.com/api3dao/airnode-protocol-v1/commit/494c66bdbe249dc3847d1379ce8d81413175a969

alikonuk1:
> I've tried using this for OEV and realized that the contract changes `msg.sender` which causes the transaction to revert, because msg.sender is checked [here](https://github.com/api3dao/airnode-protocol-v1/blob/3553a52c483525caa3afd2d244c99da75ca10335/contracts/api3-server-v1/OevDataFeedServer.sol#L66).
> 

Yes, it is true. However, as the Natspec states [here](https://github.com/api3dao/airnode-protocol-v1/blob/6a0bffbcba11c3c2da227e166aed4c91598ce541/contracts/utils/ExternalMulticallWithValue.sol#L8), this contract should only be used with functions that do not care about `msg.sender`. This PR was focused on adding a forwarding value feature. But I agree with your point, and perhaps some implementation to handle cases where `msg.sender` is important could be added.

contracts/utils/ExternalMulticall.sol
```diff
@@ -62,6 +62,59 @@ abstract contract ExternalMulticall is IExternalMulticall {
...
+        );
+        uint256 accumulatedValue = 0;
+        returndata = new bytes[](callCount);
+        for (uint256 ind = 0; ind < callCount; ) {
+            accumulatedValue += values[ind];
+            require(msg.value >= accumulatedValue, "Insufficient value");
+            require(
+                targets[ind].code.length > 0,
+                "Multicall target not contract"
+            );
```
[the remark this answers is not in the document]
alikonuk1: > I wonder if we should allow `targets[ind].code.length==0` if `data[ind].length==0` and `values[ind] != 0`. I'm thinking of a use-case where the user needs to multicall a `withdraw()` function that only transfers to `msg.sender` (though this would also mean that we would need a `receive()`).

Then this also means we would have to add a sweep function with some access control. Maybe we can implement an efficient solution like this; when it is called, `externalMulticallWithValue()` updates a storage variable `owner` to the `msg.sender`. And if some value comes back to `receive()`, it can send the value back to the `owner`. Also initialize `owner` back to `address(this)` in the end of the function.
This can also be done by just forwarding any value that comes to `receive()` to the `tx.origin`.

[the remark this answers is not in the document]
bbenligiray: > Then this also means we would have to add a sweep function with some access control

That's true, we either need `externalMulticallWithValue()` to not validate `values` based on `msg.value` or we need a sweep function. The latter is probably less error prone. 

> Maybe we can implement an efficient solution like this

That sounds very hacky. `externalMulticallWithValue()` has to be behind `onlyOwner` anyway so I don't see why we don't have the sweep function behind that as well.

bbenligiray:
> this contract should only be used with functions that do not care about msg.sender

That's true, but the whole point of this contract is for searchers to use it with Api3ServerV1 and [that does care about `msg.sender`](https://github.com/api3dao/airnode-protocol-v1/blob/main/contracts/api3-server-v1/OevDataFeedServer.sol#L66). We're still discussing how to handle this (see #technical-oev) so let us put this on hold for now.

contracts/utils/ExternalMulticallWithValue.sol
```diff
@@ -0,0 +1,74 @@
...
+    /// of the batched calls reverts
+    /// @param targets Array of target addresses of batched calls
+    /// @param data Array of calldata of batched calls
+    /// @param values Array of values to send with each call
+    /// @return returndata Array of returndata of batched calls
+    function externalMulticallWithValue(
+        address[] calldata targets,
+        bytes[] calldata data,
+        uint256[] calldata values
+    ) public payable virtual override returns (bytes[] memory returndata) {
```
bbenligiray, line 32: It's a problem that this function is `public` if someone wants to extend this contract (for example with Ownable). Maybe this should be `internal` instead and IExternalMulticallWithValue shouldn't exist.

patriksimurka:
Hi Ali, could you please rebase on top of [2.3.0](https://github.com/api3dao/airnode-protocol-v1/releases/tag/v2.3.0)? We need the changes to incorporate into OEV. Thanks

alikonuk1:
> Does not have 100% test coverage

Test coverage is now at 100%

contracts/utils/OevSearcherMulticall.sol
```diff
@@ -0,0 +1,87 @@
...
+/// about who the sender is, (2) after extending it, to interact with trusted
+/// contracts (see below for details). It implements a way of batching with value,
+/// and requires none of the calls to revert.
+/// @dev As mentioned above, this contract can be used to interact with trusted
+/// contracts. Such interactions can leave this contract in a privileged
+/// position (e.g., OevSearcherMulticall may be left with a non-zero balance of an
+/// ERC20 token as a result of a transaction sent to it), which can be abused
+/// by an attacker afterwards. In addition, attackers can frontrun interactions
+/// to have the following interaction result in an unintended outcome. A
+/// general solution to these attacks is overriding the externalMulticallWithValue function
```
bbenligiray, line 19: As a note for the future, [we limit the width to 80 characters](https://github.com/api3dao/airnode-protocol-v1/blob/main/.prettierrc.js#L18) but there is no nice way of doing this automatically for NatSpec lines so I format these manually. This line here is too long.

bbenligiray:
@alikonuk1 We need this contract soon so I pushed the changes that I want myself to buy some time. In short, I decided that limiting the external call values by `msg.value` will potentially be too restrictive (imagine that the searcher needs to borrow a flash loan of 1000 ETH and use that in their external calls, the previous implementation didn't allow that). Once I removed the check for `msg.value`, the withdrawal function becomes redundant because the owner can simply call an EOA with an arbitrary value to withdraw funds from the contract.

bbenligiray (approved):
Give it a final look and merge it

contracts/utils/OevSearcherMulticallV1.sol
```diff
@@ -0,0 +1,70 @@
...
+        uint256 callCount = targets.length;
+        require(
+            callCount == data.length && callCount == values.length,
+            "Parameter length mismatch"
+        );
+        returndata = new bytes[](callCount);
+        for (uint256 ind = 0; ind < callCount; ) {
+            bool success;
+            // solhint-disable-next-line avoid-low-level-calls
+            (success, returndata[ind]) = targets[ind].call{value: values[ind]}(
```
alikonuk1, line 40: @bbenligiray Minor but we can cut off 35 gas by caching this

bbenligiray:
```
(success, returndata[ind]) = targets[ind].call{value: values[ind]}(data[ind]);
```
is cheaper than 
```
uint256 value = values[ind];
(success, returndata[ind]) = targets[ind].call{value: value}(data[ind]);
```
The latter is only worthwhile if we use `value` again, and here, it's only used again on the non-happy path because a call failed. So changing this would be optimizing the non-happy path at the expense of the happy path, which is not something we want to do (and the same argument applies to `returndata[ind]`).

PR #133, opened 2023-05-22
Add AirnodeProtocolDryRun contract

### LINKED ISSUE #127: Add a dry run fulfillment function to AirnodeProtocol.sol
See https://github.com/api3dao/airnode/issues/1716

Since AirnodeProtocol is not deployed yet, we can add the dry run function in it to avoid having to keep track of another protocol contract address

alikonuk1:
Closes #127


bbenligiray:
The related issue was stale

PR #135, opened 2023-05-25
Deploy order payable

bbenligiray:


deploy/1_deploy.js
```diff
@@ -76,5 +81,15 @@ module.exports = async ({ getUnnamedAccounts, deployments }) => {
     });
     log(`Deployed RequesterAuthorizerWithErc721 at ${requesterAuthorizerWithErc721.address}`);
   }
+
+  if (chainsSupportedByApi3Market.includes(network.name)) {
+    const orderPayable = await deploy('OrderPayable', {
+      from: accounts[0],
+      args: [accessControlRegistry.address, 'OrderPayable admin (API3 Market)', ownableCallForwarder.address],
```
bbenligiray, line 88: Note that the admin role description is `OrderPayable admin (API3 Market)` and not `OrderPayable admin`. This is assuming that we may want to have other OrderPayable instances (and I've done the same thing for PrepaymentDepository).

aquarat (approved):
LGTM 🔥 
So the manager-multisig will have to call `initializeRoleAndGrantToSender` for the signer role, then `grantRole` for our signer EOA and then finally/technically `renounceRole` for the manager-multisig?

acenolaza:
> LGTM fire So the manager-multisig will have to call `initializeRoleAndGrantToSender` for the signer role, then `grantRole` for our signer EOA and then finally/technically `renounceRole` for the manager-multisig?

Good question but since `OwnableCallForwarder` is already set as manager I guess we might have to just call `grantRole()` with the signer EOA, right?

bbenligiray:
>  So the manager-multisig will have to call initializeRoleAndGrantToSender for the signer role, then grantRole for our signer EOA and then finally/technically renounceRole for the manager-multisig?

manager-multisig will have to call (through OwnableCallForwarder) initializeRoleAndGrantToSender for the admin role and then the signer role, then grantRole to the signer EOA. We hadn't have OwnableCallForwarder renounce the self-granted roles before and we don't need to do this here (because it's already immutably the manager and the manager can do everything that these roles can do). The unit tests renounce roles only to be able to verify that the manager can do privileged actions without self-granting the respective roles.

> Good question but since OwnableCallForwarder is already set as manager I guess we might have to just call grantRole() with the signer EOA, right?

Roles need to be initialized before you can call grantRole() for them so no, initializeRoleAndGrantToSender() will need to be called first.

PR #136, opened 2023-05-30
Update commentary

alikonuk1:
The commented function name seems to have been changed later


bbenligiray:
That's cool but changing this after deployment will change the bytecode metadata and cause contract verification to fail (see https://github.com/api3dao/airnode/blob/master/packages/airnode-protocol/scripts/verify-local.ts#L27). I'm not sure if it's worth it.

bbenligiray:
Created https://github.com/api3dao/airnode-protocol-v1/issues/138

Issue #137, opened 2023-05-31
Deploy RequesterAuthorizerWithErc721 on an additional testnet


dcroote:
Currently, `RequesterAuthorizerWithErc721` is deployed on two chains: ethereum mainnet and goerli testnet https://github.com/api3dao/airnode-protocol-v1/blob/f22a312b415bbebb287150460e3575a33193f5a3/deployments/references.json#L165-L168

However, to test the cross-chain functionality of `crossChainRequesterAuthorizersWithErc721` ([docs link here](https://docs.api3.org/reference/airnode/latest/concepts/authorizers.html#cross-chain-crosschainrequesterauthorizerswitherc721)), it would be very useful to have the contract deployed on a second testnet, for example Sepolia.

PR #139, opened 2023-05-31
Deploy RequesterAuthorizerWithErc721 on Sepolia

bbenligiray:
Closes https://github.com/api3dao/airnode-protocol-v1/issues/137

Should I publish after this so that the test uses the default address?


deploy/1_deploy.js
```diff
@@ -72,7 +72,7 @@ module.exports = async ({ getUnnamedAccounts, deployments }) => {
     log(`Deployed PrepaymentDepository (OEV Relay) at ${prepaymentDepository.address}`);
   }
 
-  if (chainsSupportedByChainApi.includes(network.name)) {
+  if (chainsSupportedByChainApi.includes(network.name) || network.name === 'ethereum-sepolia-testnet') {
```
bbenligiray, line 75: The alternative is to keep a list of chains per contract but it feels like that will be more error-prone (we will copy-paste an old version of the list to the new contract, etc.)

dcroote: I think this hardcoding is fine given `RequesterAuthorizerWithErc721` isn't meant to be deployed on a large number of chains (as opposed to say something like `AirnodeRrpV0`). My only other thought was that there could be another export below here:
https://github.com/api3dao/airnode-protocol-v1/blob/f22a312b415bbebb287150460e3575a33193f5a3/src/supported-chains.js#L16

named something like `additionalRequesterAuthorizerWithErc721Chains` that complements `chainsSupportedByChainApi`, resulting in the conditional:
```ts
if (chainsSupportedByChainApi.includes(network.name) || additionalRequesterAuthorizerWithErc721Chains.includes(network.name))
```

dcroote (approved):
Looks good- with one comment on a possible alternative to hardcoding `ethereum-sepolia-testnet` in the conditional (but feel free to ignore).

As for publishing- that would be great if it isn't too much trouble. Being able to bump the imported `airnode-protocol-v1` version that contains this additional deployment before Airnode v0.12 release would be advantageous for testing and being able to omit the contract address.

bbenligiray:
I don't want to keep contract-specific chain lists because that's hardcoding a chain list in the deployment scripts with an extra step. Maybe what's being hinted here is that whenever we deploy a contract, we should also deploy it on `ethereum-goerli-testnet` and `ethereum-sepolia-testnet` with no exceptions. I'll keep that in mind and implement it if I see this issue ever coming up again.

PR #143, opened 2023-06-07
Deploy all contracts on both Ethereum testnets

bbenligiray:
This is done to provide the dAPIs team with a testnet OrderPayable

@Ashar2shahid I couldn't deploy PrepaymentDepository on testnets because there is no mock USDC contract address. Maybe that contract should go in this repo.


vponline:
> I assume this doesn't need to be published now, let me know if it does

No big rush, I can set one of those manually for staging if the Market is ready for testing before this is published

Ashar2shahid:
don't think we need to have the token in the repo imo

PR #149, opened 2023-07-13
Deploy dAPI contracts on new testnets

bbenligiray:


bbenligiray:
Confirmed that Nodary self-funded feeds are up on these chains

PR #165, opened 2023-10-02
Adds TimestampedHashRegistry contract

acenolaza:
Closes [#102](https://github.com/api3dao/tasks/issues/102)

### DIFF
--- a/contracts/utils/TimestampedHashRegistry.sol
+++ b/contracts/utils/TimestampedHashRegistry.sol
@@ -96,8 +96,7 @@ contract TimestampedHashRegistry is
         emit RegisteredSignedHash(
             typeName,
             signedHash.hash,
-            signedHash.timestamp,
-            signatures
+            signedHash.timestamp
         );
     }
 }
--- a/contracts/utils/TimestampedHashRegistry.sol
+++ b/contracts/utils/TimestampedHashRegistry.sol
@@ -23,50 +23,50 @@ contract TimestampedHashRegistry is
 
     bytes32 private constant _SIGNED_HASH_TYPE_HASH =
         keccak256(
-            "SignedHash(bytes32 typeName,bytes32 hash,uint256 timestamp)"
+            "SignedHash(bytes32 hashType,bytes32 hash,uint256 timestamp)"
         );
 
     constructor() EIP712("TimestampedHashRegistry", "1.0.0") {}
 
-    function addSigner(bytes32 typeName, address signer) external onlyOwner {
-        require(typeName != bytes32(0), "Type name is zero");
+    function addSigner(bytes32 hashType, address signer) external onlyOwner {
+        require(hashType != bytes32(0), "Hash type is zero");
         require(signer != address(0), "Signer is zero");
         require(
-            _hashTypeToSigners[keccak256(abi.encodePacked(typeName))].add(
+            _hashTypeToSigners[keccak256(abi.encodePacked(hashType))].add(
                 signer
             ),
             "Signer already exists"
         );
-        emit AddedSigner(typeName, signer);
+        emit AddedSigner(hashType, signer);
     }
 
-    function removeSigner(bytes32 typeName, address signer) external onlyOwner {
-        require(typeName != bytes32(0), "Type name is zero");
+    function removeSigner(bytes32 hashType, address signer) external onlyOwner {
+        require(hashType != bytes32(0), "Hash type is zero");
         require(signer != address(0), "Signer is zero");
         require(
-            _hashTypeToSigners[keccak256(abi.encodePacked(typeName))].remove(
+            _hashTypeToSigners[keccak256(abi.encodePacked(hashType))].remove(
                 signer
             ),
             "Signer does not exist"
         );
-        emit RemovedSigner(typeName, signer);
+        emit RemovedSigner(hashType, signer);
     }
 
     function getSigners(
-        bytes32 typeName
+        bytes32 hashType
     ) external view returns (address[] memory signers) {
-        signers = _hashTypeToSigners[keccak256(abi.encodePacked(typeName))]
+        signers = _hashTypeToSigners[keccak256(abi.encodePacked(hashType))]
             .values();
     }
 
     function registerSignedHash(
-        bytes32 typeName,
+        bytes32 hashType,
         SignedHash calldata signedHash,
         bytes[] calldata signatures
     ) external {
-        require(typeName != bytes32(0), "Type name is zero");
+        require(hashType != bytes32(0), "Hash type is zero");
         EnumerableSet.AddressSet storage signers = _hashTypeToSigners[
-            keccak256(abi.encodePacked(typeName))
+            keccak256(abi.encodePacked(hashType))
         ];
         require(signers.length() != 0, "Signers have not been set");
         require(
@@ -80,7 +80,7 @@ contract TimestampedHashRegistry is
                         keccak256(
                             abi.encode(
                                 _SIGNED_HASH_TYPE_HASH,
-                                typeName,
+                                hashType,
                                 signedHash.hash,
                                 signedHash.timestamp
                             )
@@ -91,10 +91,10 @@ contract TimestampedHashRegistry is
             );
         }
         hashTypeToSignedHash[
-            keccak256(abi.encodePacked(typeName))
+            keccak256(abi.encodePacked(hashType))
         ] = SignedHash(signedHash.hash, signedHash.timestamp);
         emit RegisteredSignedHash(
-            typeName,
+            hashType,
             signedHash.hash,
             signedHash.timestamp
         );
--- a/contracts/utils/TimestampedHashRegistry.sol
+++ b/contracts/utils/TimestampedHashRegistry.sol
@@ -19,7 +19,9 @@ contract TimestampedHashRegistry is
 
     mapping(bytes32 => EnumerableSet.AddressSet) private _hashTypeToSigners;
 
-    mapping(bytes32 => SignedHash) public hashTypeToSignedHash;
+    mapping(bytes32 => bytes32) public hashTypeToHash;
+
+    mapping(bytes32 => uint256) public hashTypeToTimestamp;
 
     bytes32 private constant _SIGNED_HASH_TYPE_HASH =
         keccak256(
@@ -32,9 +34,7 @@ contract TimestampedHashRegistry is
         require(hashType != bytes32(0), "Hash type is zero");
         require(signer != address(0), "Signer is zero");
         require(
-            _hashTypeToSigners[keccak256(abi.encodePacked(hashType))].add(
-                signer
-            ),
+            _hashTypeToSigners[hashType].add(signer),
             "Signer already exists"
         );
         emit AddedSigner(hashType, signer);
@@ -44,9 +44,7 @@ contract TimestampedHashRegistry is
         require(hashType != bytes32(0), "Hash type is zero");
         require(signer != address(0), "Signer is zero");
         require(
-            _hashTypeToSigners[keccak256(abi.encodePacked(hashType))].remove(
-                signer
-            ),
+            _hashTypeToSigners[hashType].remove(signer),
             "Signer does not exist"
         );
         emit RemovedSigner(hashType, signer);
@@ -55,19 +53,17 @@ contract TimestampedHashRegistry is
     function getSigners(
         bytes32 hashType
     ) external view returns (address[] memory signers) {
-        signers = _hashTypeToSigners[keccak256(abi.encodePacked(hashType))]
-            .values();
+        signers = _hashTypeToSigners[hashType].values();
     }
 
-    function registerSignedHash(
+    function registerHash(
         bytes32 hashType,
-        SignedHash calldata signedHash,
+        bytes32 hash,
+        uint256 timestamp,
         bytes[] calldata signatures
     ) external {
         require(hashType != bytes32(0), "Hash type is zero");
-        EnumerableSet.AddressSet storage signers = _hashTypeToSigners[
-            keccak256(abi.encodePacked(hashType))
-        ];
+        EnumerableSet.AddressSet storage signers = _hashTypeToSigners[hashType];
         require(signers.length() != 0, "Signers have not been set");
         require(
             signatures.length == signers.length(),
@@ -80,21 +76,16 @@ contract TimestampedHashRegistry is
                         abi.encode(
                             _SIGNED_HASH_TYPE_HASH,
                             hashType,
-                            signedHash.hash,
-                            signedHash.timestamp
+                            hash,
+                            timestamp
                         )
                     )
                 ).recover(signatures[ind]) == signers.at(ind),
                 "Signature mismatch"
             );
         }
-        hashTypeToSignedHash[
-            keccak256(abi.encodePacked(hashType))
-        ] = SignedHash(signedHash.hash, signedHash.timestamp);
-        emit RegisteredSignedHash(
-            hashType,
-            signedHash.hash,
-            signedHash.timestamp
-        );
+        hashTypeToHash[hashType] = hash;
+        hashTypeToTimestamp[hashType] = timestamp;
+        emit RegisteredHash(hashType, hash, timestamp);
     }
 }
--- a/contracts/utils/TimestampedHashRegistry.sol
+++ b/contracts/utils/TimestampedHashRegistry.sol
@@ -88,13 +88,18 @@ contract TimestampedHashRegistry is
         bytes[] calldata signatures
     ) external override {
         require(hashType != bytes32(0), "Hash type is zero");
-        EnumerableSet.AddressSet storage signers = _hashTypeToSigners[hashType];
-        require(signers.length() != 0, "Signers have not been set");
         require(
-            signatures.length == signers.length(),
+            timestamp > hashTypeToTimestamp[hashType],
+            "Timestamp is not newer"
+        );
+        EnumerableSet.AddressSet storage signers = _hashTypeToSigners[hashType];
+        uint256 signersCount = signers.length();
+        require(signersCount != 0, "Signers have not been set");
+        require(
+            signatures.length == signersCount,
             "Invalid number of signatures"
         );
-        for (uint256 ind = 0; ind < signers.length(); ind++) {
+        for (uint256 ind = 0; ind < signersCount; ind++) {
             require(
                 _hashTypedDataV4(
                     keccak256(
--- a/contracts/utils/TimestampedHashRegistry.sol
+++ b/contracts/utils/TimestampedHashRegistry.sol
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.17;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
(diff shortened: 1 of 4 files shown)


contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+    mapping(bytes32 => SignedHash) public hashTypeToSignedHash;
+
+    bytes32 private constant _SIGNED_HASH_TYPE_HASH =
+        keccak256(
+            "SignedHash(bytes32 hashType,bytes32 hash,uint256 timestamp)"
+        );
+
+    constructor() EIP712("TimestampedHashRegistry", "1.0.0") {}
+
+    function setHashTypeSigners(
```
bbenligiray, line 25: The only point to use `EnumerableSet.AddressSet` for signers would be to have `addSigner()`/`removeSigner()` functions. If we're going to have to set/reset all signers, keeping them in `address[]` as in DapiFallbackV1 is much cheaper.

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+        bytes32 hashType,
+        address[] calldata signers
+    ) external onlyOwner {
+        require(hashType != bytes32(0), "Hash is zero");
+        require(signers.length != 0, "Signers length is empty");
+        for (uint256 ind = 0; ind < signers.length; ind++) {
+            _hashTypeToSigners[hashType].add(signers[ind]);
```
bbenligiray, line 32: Say I call `setHashTypeSigners(typeA, [signerA, signerB, signerC])` and then I call `setHashTypeSigners(typeA, [signerC, signerD, signerE])`. The resulting signer set will be `[signerA, signer B, signerC, signerD, signerE]`, which I suspect is not what you intend.

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+        require(hashType != bytes32(0), "Hash type is zero");
+        EnumerableSet.AddressSet storage signers = _hashTypeToSigners[hashType];
+        require(signers.length() != 0, "Signers have not been set");
+        require(
+            signatures.length == signers.length(),
+            "Signatures length mismatch"
+        );
+        for (uint256 ind = 0; ind < signatures.length; ind++) {
+            require(
+                signers.contains(
```
bbenligiray, line 51: This is kinda expensive. The DapiFallbackV1 version which expected the signatures to be ordered correctly was a good tradeoff imo.

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+        }
+        hashTypeToSignedHash[hashType] = SignedHash(
+            signedHash.hash,
+            signedHash.timestamp
+        );
+        emit RegisteredSignedHash(
+            hashType,
+            signedHash.hash,
+            signedHash.timestamp,
+            signatures
```
bbenligiray, line 74: No need to emit signatures

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+        require(signers.length != 0, "Signers length is empty");
+        for (uint256 ind = 0; ind < signers.length; ind++) {
+            _hashTypeToSigners[hashType].add(signers[ind]);
+        }
+        emit SetHashTypeSigners(hashType, signers);
+    }
+
+    function registerSignedHash(
+        bytes32 hashType,
+        SignedHash calldata signedHash,
```
bbenligiray, line 39: I don't see why the struct is not spread (though if you remove the struct altogether based on the comment above this is no longer a concern)

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+import "./interfaces/ITimestampedHashRegistry.sol";
+
+// TODO: would SelfMulticall be useful?
+contract TimestampedHashRegistry is Ownable, EIP712, ITimestampedHashRegistry {
+    using ECDSA for bytes32;
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    mapping(bytes32 => EnumerableSet.AddressSet) private _hashTypeToSigners;
+
+    mapping(bytes32 => SignedHash) public hashTypeToSignedHash;
```
bbenligiray, line 16: Say I'm building Api3Market.sol and I need to fetch the hash of a hash type. I will have to do
```solidity
(bytes32 hash, ) = timestampedHashRegistry.hashTypeToSignedHash(hashType);
```
during which I will also have to pay for `SignedHash.timestamp` to be read from storage even though I don't need it. It's better to keep the hash and the timestamp in different mappings for this reason.

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+
+    bytes32 private constant _SIGNED_HASH_TYPE_HASH =
+        keccak256(
+            "SignedHash(bytes32 hashType,bytes32 hash,uint256 timestamp)"
```
bbenligiray, line 20: If it's `hashType`, why is it not `hashTimestamp`? Since `hashTimestamp` sounds wrong, should it be `SignedHash(bytes32 type,bytes32 hash,uint256 timestamp)`?

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: That being said, `addSigner()`/`removeSigner()` is neat (but would require SelfMulticall to be inherited)

bbenligiray (changes requested):

contracts/utils/TimestampedHashRegistry.sol
acenolaza: Right, I thought about it but then I didn't want to deviate too much from DapiFallbackV1 implemenation. I'll add these functions instead 👍🏻

contracts/utils/TimestampedHashRegistry.sol
acenolaza: > This is kinda expensive. The DapiFallbackV1 version which expected the signatures to be ordered correctly was a good tradeoff imo.

I was really proud with achieving not having to rely on the caller sending the signatures in an expected order 😅 

> btw there's nothing preventing the caller from using the same signature multiple times

I'm confused by this. Wouldn't the signature check fail if a user sends a signature for a different `root` or `timestamp`? Or are you suggesting that I check the `root` parameter is different from a previously registered root for a hash type?

contracts/utils/TimestampedHashRegistry.sol
acenolaza: Correct but wouldn't that make it more expensive when registering a new hash? (write to `hashTypeToHash` then write to `hashTypeToTimestamp`). It would also require a `view` helper function for reading both values in a single RPC call too I think

contracts/utils/TimestampedHashRegistry.sol
acenolaza: or `typeName` similar to `dapiName`?

contracts/utils/TimestampedHashRegistry.sol
acenolaza: May I ask why is this a concern? Is the issue that I'm using a `struct` type as argument and this might be more expensive that just having 2 arguments? I saw that this was a common pattern like in [here](https://github.com/api3dao/airnode-protocol-v1/blob/102-TimestampedHashRegistry/contracts/utils/ExpiringMetaTxForwarder.sol#L77) and [here](https://github.com/api3dao/omnimarket/blob/main/contracts/Omnimarket.sol#L268)

contracts/utils/TimestampedHashRegistry.sol
acenolaza: I might be also useful to have a `resetSigners()` function to clear all signers for a hash type name 🤔

bbenligiray: > How about adding a signers array to the constructor arguments and do a initial set while deploying like in [DapiFallbackV1](https://github.com/api3dao/manager-multisig/blob/main/chain/contracts/DapiFallbackV1.sol#L93)? Then one can use add/remove functions later on to modify the set.

I normally avoid that for the deterministic deployments to have the same address. It's better to have an [initalization function](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L28) behind `onlyOwner`.

> I might be also useful to have a resetSigners() function to clear all signers for a hash type name 🤔 (or a setSigners(address[] signers) that first deletes all signers and then sets based on the signers param)

If [GnosisSafe](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol) doesn't need set/reset functions, neither do we imo.

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: I mean if I have one valid signature, I can repeat it `signers.length` times to create a `signatures` array that will pass this verification. Note that a signer create multiple valid signatures of the same hash (signature malleability), which means you can't avoid this by preventing signatures from being repeated. All in all, you will have to rely on the caller sending the signatures in the expected order (or sort it for them in the contract), which means the EnumerableSet is only good for having `addSigner()`/`removeSigner()` functions.

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: The first example is trying to be similar to the OpenZeppelin example meta tx forwarder contract (which no longer exists), the second example is doing it to avoid stack size limit errors. Otherwise I avoid it if I can, because for example it's not clear if/how the manager multisig frontend supports this.

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: > Correct but wouldn't that make it more expensive when registering a new hash? (write to hashTypeToHash then write to hashTypeToTimestamp).

There may be an equal cost difference, but it's better to optimize the read than the write because it's expected to be read a lot
> It would also require a view helper function for reading both values in a single RPC call too I think (or callStatic + multicall 🤔 )

You should inherit SelfMulticall anyway

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: Reserved keyword collisions are avoided by a trailing `_` as per the style guide, but if it's actually a reserved keyword forget about it

bbenligiray: `typeName` is not correct because as I've mentioned, this is not necessarily a string or the hash of a string

contracts/utils/TimestampedHashRegistry.sol
acenolaza: Got it. Good catch 👏🏻 

An alternative that I can think of is creating a new `Enumerable.AddressSet` for all addresses derived from the `signatures`. Then loop over the `signers` set and use `contains` on the derived addresses set to check if a signature for the signer is present. This would allow the caller to send the signatures in any order but it might be more expensive.

acenolaza: [This](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L134-L135) is making me uneasy tho 🤔

bbenligiray: I think using any design other than Safe's is unnecessarily adventurous here. Keep the owners in an `AddressSet`, have [`setupSigners()`](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L28), [`addSigner()`](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L58), [`removeSigner()`](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L78) functions. [Require signatures to be correctly ordered](https://github.com/safe-global/safe-contracts/blob/main/contracts/Safe.sol#L295). The CI will ensure that [our metadata](https://github.com/api3dao/manager-multisig/blob/main/data/arbitrum-goerli-testnet/metadata.json#L2) matches the on-chain order anyway and we can use that to order the signatures accordingly (btw currently the manager-multisig CI doesn't ensure this for [Safe](https://github.com/api3dao/manager-multisig/blob/main/scripts/src/verify-metadata.ts#L34) or [DapiFallbackV1](https://github.com/api3dao/manager-multisig/blob/main/scripts/src/verify-metadata.ts#L66), this needs to be fixed).

acenolaza: > have [`setupSigners()`](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L28), [`addSigner()`](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L58), [`removeSigner()`](https://github.com/safe-global/safe-contracts/blob/main/contracts/base/OwnerManager.sol#L78) functions

`setupSigners()` should completely override the elements in the set, right? This is an external call that should probably be called after deploying the contract to initialize the set of signers.

> [Require signatures to be correctly ordered](https://github.com/safe-global/safe-contracts/blob/main/contracts/Safe.sol#L295). 

`removeSigner()` would change the order of the set since [EnumerableSet._remove()](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L89-L91) swaps the last element of the set with the one being removed 😕 (btw [signatureSplit()](https://github.com/safe-global/safe-contracts/blob/main/contracts/Safe.sol#L295) only extracts the r,s,v parts from a signature at a specified position but this does not ensure that the actual signature order against the previously set owners list, I probably didn't understand the reference tho 🤭 )

> The CI will ensure that [our metadata](https://github.com/api3dao/manager-multisig/blob/main/data/arbitrum-goerli-testnet/metadata.json#L2) matches the on-chain order anyway and we can use that to order the signatures accordingly 

Does this mean that after calling `removeSigner()` and the order of the signers changes on-chain we should expect an error in CI and then fix the signers order in our metadata?

contracts/utils/TimestampedHashRegistry.sol
acenolaza: > There may be a small difference (much less than an extra storage read), but it's better to optimize the read than the write because it's expected to be read a lot anyway

Sorry if I'm being annoying but isn't this a similar case to [reading a data feed value](https://github.com/api3dao/airnode-protocol-v1/blob/102-TimestampedHashRegistry/contracts/api3-server-v1/Api3ServerV1.sol#L33-L82)? Or is timestamp even useful on this hash registry at all?

bbenligiray: It isn't because reading a data feed value reads a struct packed in a single bytes32. Here the struct isn't packed, so you'd be paying for reading two bytes32 values even if you won't be needing the timestamp, and the timestamp is much less useful to the user because unless a hash has been overwritten by a new one it still applies, while a data feed value goes stale so the user may want to read the timestamp to check for that.

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: > owners == signers, right?

Right

> setupSigners() should completely override the elements in the set, like a reset? or should this be an external function that should only be called once for each hashType to initialize the set of signers (if signers have ever been set for the hashType then revert similar to Safe.sol [setup()](https://github.com/safe-global/safe-contracts/blob/main/contracts/Safe.sol#L106)).

The latter

> removeSigner() would change the order of the set since [EnumerableSet._remove()](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/EnumerableSet.sol#L89-L91) swaps the last element of the set with the one being removed 😕

Not a problem. It's the users' (or maybe the manager-multisig frontend's) responsibility to update the metadata accordingly.

> (btw [signatureSplit()](https://github.com/safe-global/safe-contracts/blob/main/contracts/Safe.sol#L295) only extracts the r,s,v parts from a signature at a specified position but this does not ensure the actual signature order against the previously set owners list, I probably didn't understand the reference tho 🤭 )

I mean Safe also depends on sending the signatures in the correct order, sending them in an array or in serialized form doesn't matter

> Does this mean that after calling removeSigner() and the order of the signers changes on-chain we should expect an error in CI and then fix the signers order in our metadata?

Either that or the frontend should have a "create tx to remove signer" button that you click that updates the metadata as expected and create the transaction at the same time. Once the transaction is executed the CI passes automatically.

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,116 @@
...
+        emit RemovedSigner(hashType, signer);
+    }
+
+    function getSigners(
+        bytes32 hashType
+    ) external view override returns (address[] memory signers) {
+        signers = _hashTypeToSigners[hashType].values();
+    }
+
+    function registerHash(
```
alikonuk1, line 84: This function doesn't really check if the given timestamp is the newest for a hash. So, an old timestamp can be resubmitted with the correct hash and valid signatures from the correct set of signers.

contracts/utils/TimestampedHashRegistry.sol
```diff
@@ -0,0 +1,116 @@
...
+        uint256 timestamp,
+        bytes[] calldata signatures
+    ) external override {
+        require(hashType != bytes32(0), "Hash type is zero");
+        EnumerableSet.AddressSet storage signers = _hashTypeToSigners[hashType];
+        require(signers.length() != 0, "Signers have not been set");
+        require(
+            signatures.length == signers.length(),
+            "Invalid number of signatures"
+        );
```
alikonuk1, line 96:
```suggestion
        );
        require(timestamp > hashTypeToTimestamp[hashType], "Timestamp is not newer");
```

alikonuk1 (changes requested):

contracts/utils/TimestampedHashRegistry.sol
acenolaza: Applying the same logic, should we also check that hash is != to hash on chain?

alikonuk1: Makes sense since I cant think of any case where we would need to reregister the same hash with a newer timestamp.

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: Oof good catch

contracts/utils/TimestampedHashRegistry.sol
bbenligiray: I can think of one scenario. Say we have a valid signed hash that we have registered. Then we sign another hash with a more recent timestamp. Before registering it, we realize that the new hash is wrong and we want to keep the old one. We would want to update with the same hash with an even more recent timestamp to invalidate the wrong hash so that it can't be registered by anyone.

alikonuk1 (approved):

acenolaza:
I don't care about tests being broken by https://github.com/api3dao/airnode-protocol-v1/pull/165/commits/dfb4ae92e6b0daca81ee9ac96afba47116e676f5 because this contract will be moved to @api3/dapi-management repo where solidity compiler will be set to 0.8.18

acenolaza:
Closing this in favor of [this PR](https://github.com/api3dao/dapi-management/pull/1)

